PR #15, opened 2024-06-25
Cleanup microcluster/k8s when scaling down cluster nodes

bschimke95:
## Summary

Remove nodes from microcluster and k8s when scaling down the cluster.

## Notable changes

* Moved k8sd-proxy requests into generic separate function
* Added k8sd-proxy blacklist options to ensure the proxy is not targeting the machine that is removed
* Added `RemoveMachineFromCluster` method to workload cluster

## Tests

Manual tests for now:
* Created 4 CP node cluster - removed 1 of them again
  -> verified on remaining nodes that node is removed from microcluster and roles are updated
* Created 1 CP, 1 worker cluster - removed worker
  -> verified that worker is removed from k8s

### DIFF
--- a/bootstrap/controllers/ck8sconfig_controller.go
+++ b/bootstrap/controllers/ck8sconfig_controller.go
@@ -298,9 +298,7 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	// Accept any hostname by passing an empty string
-	// Some infrastructures will have machines where hostname and machine name do not match by design (e.g. AWS)
-	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, "")
+	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to request join token: %w", err)
 	}
--- a/bootstrap/controllers/ck8sconfig_controller.go
+++ b/bootstrap/controllers/ck8sconfig_controller.go
@@ -235,7 +235,7 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	joinToken, err := workloadCluster.NewControlPlaneJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, machine.Name)
+	joinToken, err := workloadCluster.NewControlPlaneJoinToken(ctx, *authToken, scope.Config.Name)
 	if err != nil {
 		return fmt.Errorf("failed to request join token: %w", err)
 	}
@@ -308,7 +308,9 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, machine.Name)
+	// Accept any hostname by passing an empty string
+	// Some infrastructures will have machines where hostname and machine name do not match by design (e.g. AWS)
+	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, *authToken, "")
 	if err != nil {
 		return fmt.Errorf("failed to request join token: %w", err)
 	}
--- a/bootstrap/controllers/ck8sconfig_controller.go
+++ b/bootstrap/controllers/ck8sconfig_controller.go
@@ -235,7 +235,7 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	joinToken, err := workloadCluster.NewControlPlaneJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, scope.Config.Name)
+	joinToken, err := workloadCluster.NewControlPlaneJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, machine.Name)
 	if err != nil {
 		return fmt.Errorf("failed to request join token: %w", err)
 	}
@@ -308,7 +308,7 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, scope.Config.Name)
+	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, machine.Name)
 	if err != nil {
 		return fmt.Errorf("failed to request join token: %w", err)
 	}
--- a/bootstrap/controllers/ck8sconfig_controller.go
+++ b/bootstrap/controllers/ck8sconfig_controller.go
@@ -220,7 +220,7 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 	// injects into config.Version values from top level object
 	r.reconcileTopLevelObjectSettings(scope.Cluster, machine, scope.Config)
 
-	workloadCluster, err := r.managementCluster.GetWorkloadCluster(ctx, util.ObjectKey(scope.Cluster), scope.Config.Spec.ControlPlaneConfig.MicroclusterPort)
+	workloadCluster, err := r.managementCluster.GetWorkloadCluster(ctx, util.ObjectKey(scope.Cluster), scope.Config.Spec.ControlPlaneConfig.GetMicroclusterPort())
 	if err != nil {
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
@@ -293,7 +293,7 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
 		return fmt.Errorf("auth token not yet generated")
 	}
 
-	workloadCluster, err := r.managementCluster.GetWorkloadCluster(ctx, util.ObjectKey(scope.Cluster), scope.Config.Spec.ControlPlaneConfig.MicroclusterPort)
+	workloadCluster, err := r.managementCluster.GetWorkloadCluster(ctx, util.ObjectKey(scope.Cluster), scope.Config.Spec.ControlPlaneConfig.GetMicroclusterPort())
 	if err != nil {
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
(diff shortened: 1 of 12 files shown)


bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -235,7 +235,7 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	joinToken, err := workloadCluster.NewControlPlaneJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, scope.Config.Name)
+	joinToken, err := workloadCluster.NewControlPlaneJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, machine.Name)
```
bschimke95, line 238: drive-by fix: the worker hostname and token name did not match

controlplane/controllers/scale.go
```diff
@@ -159,6 +160,29 @@ func (r *CK8sControlPlaneReconciler) scaleDownControlPlane(
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to lookup auth token: %w", err)
+	}
+
+	if authToken == nil {
+		return ctrl.Result{}, fmt.Errorf("auth token not yet generated")
+	}
+
+	// TODO: Get from CK8sConfig
+	microclusterPort := 2380
```
bschimke95, line 180: I could not really figure out what is the best way to access the cluster config from this controller.

controlplane/main.go
```diff
@@ -66,7 +66,7 @@ func main() {
 	flag.DurationVar(&syncPeriod, "sync-period", 10*time.Minute,
 		"The minimum interval at which watched resources are reconciled (e.g. 15m)")
 
-	flag.DurationVar(&k8sdDialTimeout, "k8sd-dial-timeout-duration", 10*time.Second,
+	flag.DurationVar(&k8sdDialTimeout, "k8sd-dial-timeout-duration", 60*time.Second,
```
bschimke95, line 69: Currently, the node-removal k8s-snap endpoint waits until the node is removed. This may take more than 10s

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -308,7 +308,7 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
 		return fmt.Errorf("failed to create remote cluster client: %w", err)
 	}
 
-	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, scope.Config.Name)
+	joinToken, err := workloadCluster.NewWorkerJoinToken(ctx, *authToken, scope.Config.Spec.ControlPlaneConfig.MicroclusterPort, machine.Name)
```
neoaggelos, line 311: TLDR: for worker nodes, please use an empty string instead `""`, such that any hostname is accepted. 

This is very important for worker nodes, as different infrastructures will have machines where hostname and machine name do not match by design.

for example, in AWS, machine names will be `$cluster-control-plane-$foobar`, and then hostnames will be `ip-$1-$2-$3-$4`. this is something we _need_, since the external cloud providers sometimes necessitate that the node matches specific criteria, see https://cloud-provider-aws.sigs.k8s.io/prerequisites/#node-name-conventions

controlplane/controllers/scale.go
neoaggelos: NOTE: microcluster port is not something that changes, and therefore should be handled in the workload cluster, not in every method.

controlplane/controllers/machine_controller.go
```diff
@@ -36,15 +34,15 @@ func (r *MachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
...
 
 	// NOTE(neoaggelos): See note below
-	/**
 	if r.managementCluster == nil {
 		r.managementCluster = &ck8s.Management{
-			Client:          r.Client,
-			EtcdDialTimeout: r.EtcdDialTimeout,
-			EtcdCallTimeout: r.EtcdCallTimeout,
+			Client: r.Client,
+			/*
```
neoaggelos, line 40: let's set the dial timeout here

controlplane/controllers/scale.go
```diff
@@ -159,6 +160,27 @@ func (r *CK8sControlPlaneReconciler) scaleDownControlPlane(
...
+	}
+
+	authToken, err := token.Lookup(ctx, r.Client, clusterObjectKey)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to lookup auth token: %w", err)
+	}
+
+	if authToken == nil {
+		return ctrl.Result{}, fmt.Errorf("auth token not yet generated")
+	}
```
neoaggelos, line 177: The token lookup feels a bit out of place here, maybe it could be included in the GetWorkloadCluster, or perhaps the `RemoveMachineFromCluster` call. But that's ok, let's sleep on this and see how the code evolves before any changes.

controlplane/controllers/scale.go
```diff
@@ -159,6 +160,27 @@ func (r *CK8sControlPlaneReconciler) scaleDownControlPlane(
...
+
+	err = workloadCluster.RemoveMachineFromCluster(ctx, machineToDelete, *authToken)
+	if err != nil {
```
neoaggelos, line 180: otherwise err trickles down to the rest of the code, let's avoid this

```suggestion
	if err := workloadCluster.RemoveMachineFromCluster(ctx, machineToDelete, *authToken); err != nil {
```

pkg/ck8s/workload_cluster.go
```diff
@@ -172,13 +175,24 @@ func getNodeInternalIP(node *corev1.Node) (string, error) {
 	return "", fmt.Errorf("unable to find internal IP for node %s", node.Name)
 }
 
-func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context) (*K8sdClient, error) {
+type k8sdProxyOptions struct {
+	// BlacklistedControlPlanes is a list of control plane nodes which k8sd proxy should not use.
+	// This is useful when a control plane node is being removed and we want to avoid using it
+	BlacklistedControlPlanes []string
```
neoaggelos, line 181: i'd make this a `IgnoreNodes map[string]struct{}`

pkg/ck8s/workload_cluster.go
```diff
@@ -172,13 +175,24 @@ func getNodeInternalIP(node *corev1.Node) (string, error) {
...
+
+// GetK8sdProxyForControlPlane returns a k8sd proxy client for the control plane.
+func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context, options k8sdProxyOptions) (*K8sdClient, error) {
 	cplaneNodes, err := w.getControlPlaneNodes(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get control plane nodes: %w", err)
 	}
 
 	for _, node := range cplaneNodes.Items {
+		if slices.Contains(options.BlacklistedControlPlanes, node.Name) {
```
neoaggelos, line 192: then this a `_, ok := opts.IgnoredNodes[node.Name]; ok`

pkg/ck8s/workload_cluster.go
```diff
@@ -192,58 +206,93 @@ func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context) (*K8sdClient
...
-		Error    string                     `json:"error"`
-		Metadata apiv1.GetJoinTokenResponse `json:"metadata"`
+		Error    string          `json:"error"`
+		Metadata json.RawMessage `json:"metadata"`
 	}
 
-	requestBody, err := json.Marshal(apiv1.GetJoinTokenRequest{Name: name, Worker: worker})
+	// FIXME: Get this from the workload cluster's configuration instead of hardcoding it.
+	// See https://github.com/canonical/cluster-api-k8s/pull/8#discussion_r1645536362 for context.
+	microclusterPort := 2380
```
neoaggelos, line 250: please, let's take this in when calling `GetWorkloadCluster()`, do not hardcode this.

pkg/ck8s/workload_cluster.go
```diff
@@ -192,58 +206,93 @@ func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context) (*K8sdClient
...
+
+	if response == nil {
+		// Nothing to decode
+		return nil
+	}
 
-	var wrappedResp wrappedResponse
-	if err := json.NewDecoder(res.Body).Decode(&wrappedResp); err != nil {
-		return "", fmt.Errorf("failed to parse HTTP response: %w", err)
+	respBody := wrappedResponse{}
```
neoaggelos, line 284: more idiomatic
```suggestion
	var respBody wrappedResponse
```

neoaggelos: also, since we have `requestBody` earlier, let this be `responseBody`

controlplane/controllers/scale.go
bschimke95: added it to the `GetWorkloadCluster` call

controlplane/controllers/scale.go
bschimke95: Moved it to the GetWorkloadCluster call, feels quite right there.

pkg/ck8s/workload_cluster.go
```diff
@@ -43,8 +43,10 @@ type WorkloadCluster interface {
 	ClusterStatus(ctx context.Context) (ClusterStatus, error)
 	UpdateAgentConditions(ctx context.Context, controlPlane *ControlPlane)
 	UpdateEtcdConditions(ctx context.Context, controlPlane *ControlPlane)
-	NewControlPlaneJoinToken(ctx context.Context, authToken string, microclusterPort int, name string) (string, error)
-	NewWorkerJoinToken(ctx context.Context, authToken string, microclusterPort int, name string) (string, error)
+	NewControlPlaneJoinToken(ctx context.Context, name string) (string, error)
+	NewWorkerJoinToken(ctx context.Context, name string) (string, error)
```
neoaggelos, line 47: final note, let's drop the `name` from the interface here. We expect this to always be `""` so let's hardcode it into the implementation. then callers don't need to care.

_If we ever change this_, we'll add a V2, or change the interface, then callers must adapt and we don't miss any required updates in the code.

pkg/ck8s/workload_cluster.go
```diff
@@ -67,10 +69,11 @@ type WorkloadCluster interface {
 // Workload defines operations on workload clusters.
 type Workload struct {
 	WorkloadCluster
-
+	AuthToken           string
 	Client              ctrlclient.Client
 	ClientRestConfig    *rest.Config
 	K8sdClientGenerator *k8sdClientGenerator
+	MicroclusterPort    int
```
neoaggelos, line 76: any objections to making `MicroclusterPort` and `AuthToken` private fields?

pkg/ck8s/workload_cluster.go
```diff
@@ -192,58 +206,91 @@ func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context) (*K8sdClient
...
-	if res.StatusCode != 200 {
-		return "", fmt.Errorf("request for join token request failed: %s", wrappedResp.Error)
+
+	var responseBody wrappedResponse
+	if err := json.NewDecoder(res.Body).Decode(&responseBody); err != nil {
+		return fmt.Errorf("failed to parse HTTP response: %w", err)
+	}
+	if res.StatusCode != http.StatusOK {
+		return fmt.Errorf("HTTP request failed with status code: %d (%s)", res.StatusCode, responseBody.Error)
+	}
```
neoaggelos, line 285: should we not first check the status coe, then start decoding things (or skip decoding if the response is `nil`)?

neoaggelos (approved):
LGTM!

pkg/ck8s/management_cluster.go
```diff
@@ -87,10 +88,25 @@ func (m *Management) GetWorkloadCluster(ctx context.Context, clusterKey client.O
...
+		return nil, fmt.Errorf("failed to lookup auth token: %w", err)
+	}
+
+	if authToken == nil {
+		return nil, fmt.Errorf("auth token not yet generated")
+	}
+
+	if microclusterPort == 0 {
+		microclusterPort = 2380
+	}
```
neoaggelos, line 102: the more i see this, the more i think we should approach this differently.

Make the MicroclusterPort field a `*int`, then add a `GetMicroclusterPort()` method, which returns the field (if set), otherwise the default value (in this case 2380)`

then use `.GetMicroclusterPort()` for all invocations and remove this.

bschimke95: done

PR #16, opened 2024-06-25
Add unit tests for `cloudinit` generation

mateoflorido:
## Overview
Add unit tests for the `cloudinit` package.
## Rationale
These unit tests aim to validate the following aspects of cloudinit configurations:
- Expected commands within each config.
 - Expected files within configs for:
   - Initializing control planes
   - Joining new control planes
   - Joining new workers

### DIFF
--- a/pkg/cloudinit/controlplane_init_test.go
+++ b/pkg/cloudinit/controlplane_init_test.go
@@ -68,7 +68,7 @@ func TestNewInitControlPlane(t *testing.T) {
 	}))
 
 	// NOTE (mateoflorido): Keep this test in sync with the expected paths in the controlplane_init.go file.
-	expectedPaths := []interface{}{
+	g.Expect(config.WriteFiles).To(ConsistOf(
 		HaveField("Path", "/capi/scripts/install.sh"),
 		HaveField("Path", "/capi/scripts/bootstrap.sh"),
 		HaveField("Path", "/capi/scripts/load-images.sh"),
@@ -83,9 +83,7 @@ func TestNewInitControlPlane(t *testing.T) {
 		HaveField("Path", "/capi/etc/snap-track"),
 		HaveField("Path", "/capi/manifests/00-k8sd-proxy.yaml"),
 		HaveField("Path", "/tmp/file"),
-	}
-
-	g.Expect(config.WriteFiles).To(ConsistOf(expectedPaths...), "Some /capi/scripts files are missing")
+	), "Some /capi/scripts files are missing")
 }
 
 func TestNewInitControlPlaneInvalidVersionError(t *testing.T) {
--- a/pkg/cloudinit/controlplane_init_test.go
+++ b/pkg/cloudinit/controlplane_init_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cloudinit_test
 
 import (
-	"os"
 	"testing"
 
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


pkg/cloudinit/controlplane_init_test.go
```diff
@@ -40,23 +41,127 @@ func TestNewInitControlPlane(t *testing.T) {
...
+	scriptFiles := map[string]string{
+		"./scripts/install.sh":                   "/capi/scripts/install.sh",
+		"./scripts/bootstrap.sh":                 "/capi/scripts/bootstrap.sh",
+		"./scripts/load-images.sh":               "/capi/scripts/load-images.sh",
+		"./scripts/join-cluster.sh":              "/capi/scripts/join-cluster.sh",
+		"./scripts/wait-apiserver-ready.sh":      "/capi/scripts/wait-apiserver-ready.sh",
+		"./scripts/deploy-manifests.sh":          "/capi/scripts/deploy-manifests.sh",
+		"./scripts/configure-token.sh":           "/capi/scripts/configure-token.sh",
+		"./scripts/create-sentinel-bootstrap.sh": "/capi/scripts/create-sentinel-bootstrap.sh",
+	}
```
mateoflorido, line 120: By default, `NewBaseCloudConfig` dumps all the scripts in the WriteFiles. Should we only assert that the necessary scripts are included in the `config.WriteFiles`?

pkg/cloudinit/controlplane_join_test.go
```diff
@@ -42,4 +48,99 @@ func TestNewJoinControlPlane(t *testing.T) {
...
+			Content:     string(content),
+			Permissions: "0500",
+			Owner:       "root:root",
+		})
+	}
+
+	g.Expect(config.WriteFiles).To(ConsistOf(expectedWriteFiles))
+}
+
+func TestNewJoinControlPlaneError(t *testing.T) {
```
bschimke95, line 112: nit: name this something like `TestNewJoinControlPlaneInvalidVersionError` otherwise I have to lookup the implementation on what kind of error this checks.

pkg/cloudinit/controlplane_init_test.go
bschimke95: It's fine to just assert that the script files exist.

bschimke95 (commented):
Great work @mateoflorido only 2 minor comments

pkg/cloudinit/controlplane_init_test.go
mateoflorido: I think we can keep it like this. This will allow us to discover if, for any reason, the content, permissions, or owner of the script are different. What do you think?

bschimke95: Fair enough :)

bschimke95 (approved):
LGTM

pkg/cloudinit/controlplane_init_test.go
neoaggelos, line 120: this test does not protect us from any regressions. as such, it's best to write it differently (e.g. that there is a cloudinit.File with Path /capi/scripts/install.sh, bootstrap.sh, etc.

As written, this is just something that we will be updating along with the implementation, so offers very little value

this is okay but we are not really testing anything with it. it is enough to check that the files are in the writeFiles.


this is because, if we ever add any new files, this test breaks by definition. it does not prevent any regressions

neoaggelos (commented):
Looks good, but left a comment for the check on the capi scripts

pkg/cloudinit/controlplane_join_test.go
```diff
@@ -42,4 +47,59 @@ func TestNewJoinControlPlane(t *testing.T) {
...
+		HaveField("Path", "/capi/scripts/configure-token.sh"),
+		HaveField("Path", "/capi/scripts/create-sentinel-bootstrap.sh"),
+		HaveField("Path", "/capi/etc/config.yaml"),
+		HaveField("Path", "/capi/etc/microcluster-address"),
+		HaveField("Path", "/capi/etc/join-token"),
+		HaveField("Path", "/capi/etc/snap-track"),
+		HaveField("Path", "/tmp/file"),
+	}
+
+	g.Expect(config.WriteFiles).To(ConsistOf(expectedPaths...), "Some /capi/scripts files are missing")
```
neoaggelos, line 68: better inline the expectedPaths here, also in other places

neoaggelos (approved):
just a final nit then we're good to go

PR #18, opened 2024-07-02
Add Go CI

bschimke95:
Runs the Go formatting and tests as part of the CI.

Also fixes formatting issues that popped up when running the CI.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -1,0 +1,42 @@
+name: Go
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    permissions:
+      contents: read  # for actions/checkout to fetch code
+    name: Unit Tests & Code Quality
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@v2
+        with:
+          egress-policy: audit
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Install Go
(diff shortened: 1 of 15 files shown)


.github/workflows/go.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    permissions:
+      contents: read  # for actions/checkout to fetch code
```
neoaggelos, line 15: is this different from the top-level line 9?

.github/workflows/go.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+
+      - name: go fmt
+        run: make go-fmt
+
+      - name: go vet
+        run: make go-vet
```
neoaggelos, line 36: also add a step for `make lint`

neoaggelos (commented):
forgot to submit earlier

Makefile
```diff
@@ -177,6 +177,17 @@ BOOTSTRAP_IMG ?= $(REGISTRY)/bootstrap-controller
...
 CONTROLPLANE_IMG ?= $(REGISTRY)/controlplane-controller
 
+go-vet:
+	go vet ./...
+
+go-fmt:
+	go fmt ./...
+
+go-lint:
+	golangci-lint run
```
neoaggelos, line 187: best revert this?

neoaggelos (approved):
Final nit and we should be good to go!

PR #19, opened 2024-07-02
Add CI for provider images

bschimke95:
## Summary

Adds a CI job that builds and publishes the bootstrap/control-plane controller docker images and publishes them to `ghcr.io`. Also, a release is created (version is based on the tag) that includes the bootstrap and controlplane resources.

See https://github.com/canonical/cluster-api-k8s/releases for an example release.


.github/workflows/release.yaml
```diff
@@ -0,0 +1,69 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+'
+  pull_request:
```
bschimke95, line 8: TODO: remove before merge

neoaggelos (approved):
Revert the version tags and gh action triggers, then it's good to go! Nice work!

PR #24, opened 2024-07-08
Add MaaS cluster template

berkayoz:
Added MaaS cluster template


templates/maas/cluster-template.yaml
```diff
@@ -0,0 +1,104 @@
...
+spec:
+  machineTemplate:
+    infrastructureTemplate:
+      apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+      kind: MaasMachineTemplate
+      name: ${CLUSTER_NAME}-control-plane
+  spec:
+    controlPlane:
+      extraKubeAPIServerArgs:
+        --anonymous-auth: "true"
```
neoaggelos, line 44: is this needed? maas does dns load-balancing, so this should not be required to get a working cluster

templates/maas/cluster-template.rc
```diff
@@ -0,0 +1,25 @@
+# Kubernetes cluster configuration
+export KUBERNETES_VERSION=1.30.2
```
neoaggelos, line 2: let's be consistent in docs, use `v1.30.2`

neoaggelos (approved):

PR #25, opened 2024-07-10
vSphere infrastructure provider template

eaudetcobello:
Gaps:

- `insecure-flag: true` issue and sha1/sha256 thumbprint needs to be investigated. TLDR, there's conflicting documentation where the provider says the thumbprint needs to be in sha256 format, but providing the thumbprint in sha256 causes a "thumbprint mismatch" error. I have resolved this by using a sha1 thumbprint *and* setting `insecure-flag: true`. Need to try *only* setting `insecure-flag: true` with sha256, or removing `insecure-flag: true` and use sha1.

**NOTE** Do not merge before testing with multi-node cluster.
**NOTE** Verify if L297-301, L319-323, L293-295 are needed for multi-node cluster before merging.


templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+metadata:
+  labels:
+    cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
+  name: ${CLUSTER_NAME}
+  namespace: default
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks:
+        - 192.168.0.0/16
```
neoaggelos, line 12: let's use the same CIDRs we have in other places, `10.1.0.0/16` for pods and `10.152.183.0/24` for services

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+                  name: kube-vip
+                  resources: {}
+                  securityContext:
+                    capabilities:
+                      add:
+                      - NET_ADMIN
+                      - NET_RAW
+                  volumeMounts:
+                  - mountPath: /etc/kubernetes/admin.conf
+                    name: kubeconfig
```
neoaggelos, line 185: you should create the RBAC rules and use that serviceaccount instead of mounting the admin kubeconfig

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+    preRunCommands:
+      - hostnamectl set-hostname "{{ ds.meta_data.hostname }}"
+      - echo "::1         ipv6-localhost ipv6-loopback localhost6 localhost6.localdomain6"
+        >/etc/hosts
+      - echo "127.0.0.1   {{ ds.meta_data.hostname }} {{ local_hostname }} localhost
+        localhost.localdomain localhost4 localhost4.localdomain4" >>/etc/hosts
+      - echo "${VSPHERE_SSH_AUTHORIZED_KEY}" >> /root/.ssh/authorized_keys
+      - chmod 700 /root/.ssh
+      - chmod 600 /root/.ssh/authorized_keys
+      - mkdir -p /etc/systemd/system/snap.k8s.containerd.service.d
```
neoaggelos, line 220: we don't need this, the cloud-init files will create intermediate directories as needed

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+      - sudo snap install k8s --classic --edge
```
neoaggelos, line 221: sudo is not needed, add a NOTE why this is here (this needs to go away before merging)

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+      kind: VSphereMachineTemplate
+      name: ${CLUSTER_NAME}
+  replicas: 3
+  version: ${K8S_VERSION}
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
+kind: CK8sConfigTemplate
+metadata:
+  name: ${CLUSTER_NAME}-md-0
+  namespace: default
```
neoaggelos, line 234: from here and other resources, remove the `namespace:` (NOTE: do not remove it from the cloud-provider manifests). the namespace is decided by the `clusterctl generate cluster --target-namespace $ns` argument

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: MachineDeployment
+metadata:
+  labels:
+    cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
+  name: ${CLUSTER_NAME}-md-0
+  namespace: default
+spec:
+  clusterName: ${CLUSTER_NAME}
+  replicas: 3
```
neoaggelos, line 256: this should be ${WORKER_MACHINE_COUNT}

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+      cloudProvider: external
+    files:
+      - content: |
+          [Service]
+          Environment="HTTPS_PROXY=http://squid.internal:3128"
+          Environment="HTTP_PROXY=http://squid.internal:3128"
+          Environment="https_proxy=http://squid.internal:3128"
+          Environment="http_proxy=http://squid.internal:3128"
+          Environment="NO_PROXY=10.0.0.0/8,192.168.0.0/16,127.0.0.1,localhost,0.0.0.0"
+          Environment="no_proxy=10.0.0.0/8,192.168.0.0/16,127.0.0.1,localhost,0.0.0.0"
```
neoaggelos, line 108: two notes here:

- best move these to a secret (e.g. `${CLUSTER_NAME}-resources`) and then use `contentFrom` here to keep this section clean and manageable
- for the proxy configuration specifically, we cannot hardcode it into the template (having a vsphere environment is orthogonal to requiring a proxy). maybe you could add a ${VSPHERE_PROXY_DISABLED} prefix on all lines here, then add it to the list of vars as empty (to enable) or `# ` to disable.
-

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1457 @@
...
+      - chmod 700 /root/.ssh
+      - chmod 600 /root/.ssh/authorized_keys
+      - mkdir -p /etc/systemd/system/snap.k8s.containerd.service.d
+      - sudo snap install k8s --classic --edge
+  machineTemplate:
+    infrastructureTemplate:
+      apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+      kind: VSphereMachineTemplate
+      name: ${CLUSTER_NAME}
+  replicas: 3
```
neoaggelos, line 227: this should be ${CONTROL_PLANE_MACHINE_COUNT}

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,22 @@
...
+export VIP_NETWORK_INTERFACE="ens192"                               # The interface that kube-vip should apply the IP to. Omit to tell kube-vip to autodetect the interface.
+export VSPHERE_TLS_THUMBPRINT="..."                                 # sha256 thumbprint of the vcenter certificate: openssl x509 -sha256 -fingerprint -in ca.crt -noout
+export EXP_CLUSTER_RESOURCE_SET="true"                              # This enables the ClusterResourceSet feature that we are using to deploy CSI
+export VSPHERE_SSH_AUTHORIZED_KEY="ssh-rsa AAAAB3N..."              # The public ssh authorized key on all machines in this cluster.
+                                                                    #  Set to "" if you don't want to enable SSH, or are using another solution.
+export VSPHERE_STORAGE_POLICY=""                                    # This is the vSphere storage policy. Set it to "" if you don't want to use a storage policy.
+export "CPI_IMAGE_K8S_VERSION"="v1.28.0"                            # The version of the vSphere CPI image to be used by the CPI workloads
```
neoaggelos, line 20: let's set the whole image as an argument here, same for csi

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,22 @@
+# Source: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/docs/getting_started.md
+
```
neoaggelos, line 2: add CONTROL_PLANE_MACHINE_COUNT and WORKER_MACHINE_COUNT vars

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,22 @@
...
+export VSPHERE_USERNAME="vi-admin@vsphere.local"                    # The username used to access the remote vSphere endpoint
+export VSPHERE_PASSWORD="admin!23"                                  # The password used to access the remote vSphere endpoint
```
neoaggelos, line 4: let's avoid this

neoaggelos (changes requested):
Immaculate work getting it to work! Let's clean up the rough edges!

templates/vsphere/template-variables.rc
eaudetcobello: Avoid what?

neoaggelos: Let's avoid fake credentials. This should be empty or commented out, so that if people forget to add their password, they will fail to generate a cluster template altogether (rather than getting a cluster template with a placeholder password)

templates/vsphere/template-variables.rc
eaudetcobello: Smart

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+export CSI_DRIVER_IMAGE="gcr.io/cloud-provider-vsphere/csi/release/driver:v3.3.0"
+export CSI_SYNCER_IMAGE="gcr.io/cloud-provider-vsphere/csi/release/syncer:v3.3.0"
+export CSI_LIVENESS_PROBE_IMAGE="registry.k8s.io/sig-storage/livenessprobe:v2.10.0"
+export CSI_ATTACHER_IMAGE="registry.k8s.io/sig-storage/csi-attacher:v4.3.0"
+export CSI_RESIZER_IMAGE="registry.k8s.io/sig-storage/csi-resizer:v1.8.0"
+export CSI_PROVISIONER_IMAGE="registry.k8s.io/sig-storage/csi-provisioner:v3.5.0"
+export CSI_SNAPSHOTTER_IMAGE="registry.k8s.io/sig-storage/csi-snapshotter:v6.2.2"
+export CSI_REGISTRAR_IMAGE="registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.8.0"
+
+export CK8S_VERSION="v1.30.0"                                                   # K8s Version of the cluster
```
neoaggelos, line 31: the standard name for that is `KUBERNETES_VERSION`, which comes from `clusterctl generate cluster --kubernetes-version`

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+export VSPHERE_DATASTORE="DefaultDatastore"                                     # The vSphere datastore to deploy the management cluster on
+export VSPHERE_NETWORK="VM Network"                                             # The VM network to deploy the management cluster on
+export VSPHERE_RESOURCE_POOL="*/Resources"                                      # The vSphere resource pool for your VMs
+export VSPHERE_FOLDER="vm"                                                      # The VM folder for your VMs. Set to "" to use the root vSphere folder
+export VSPHERE_TEMPLATE="ubuntu-1804-kube-v1.17.3"                              # The VM template to use for your management cluster.
+export CONTROL_PLANE_ENDPOINT_IP="192.168.9.230"                                # the IP that kube-vip is going to use as a control plane endpoint
+export VIP_NETWORK_INTERFACE="ens192"                                           # The interface that kube-vip should apply the IP to. Omit to tell kube-vip to autodetect the interface.
+export VSPHERE_TLS_THUMBPRINT="..."                                             # sha1 thumbprint of the vcenter certificate: openssl x509 -sha1 -fingerprint -in ca.crt -noout
```
neoaggelos, line 15: same here, leave these unset, and add a `, e.g. "10.0.0.1"` etc at the end of the comments

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+export CSI_PROVISIONER_IMAGE="registry.k8s.io/sig-storage/csi-provisioner:v3.5.0"
+export CSI_SNAPSHOTTER_IMAGE="registry.k8s.io/sig-storage/csi-snapshotter:v6.2.2"
+export CSI_REGISTRAR_IMAGE="registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.8.0"
+
+export CK8S_VERSION="v1.30.0"                                                   # K8s Version of the cluster
+
+export VSPHERE_PROXY_DISABLE="#"                                                # Set to "#" to disable, or "" to enable the proxy configuration
+
+export CONTROL_PLANE_MACHINE_COUNT=3
+export WORKER_MACHINE_COUNT=1
```
neoaggelos, line 36: newline at EOF

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+export CSI_LIVENESS_PROBE_IMAGE="registry.k8s.io/sig-storage/livenessprobe:v2.10.0"
+export CSI_ATTACHER_IMAGE="registry.k8s.io/sig-storage/csi-attacher:v4.3.0"
+export CSI_RESIZER_IMAGE="registry.k8s.io/sig-storage/csi-resizer:v1.8.0"
+export CSI_PROVISIONER_IMAGE="registry.k8s.io/sig-storage/csi-provisioner:v3.5.0"
+export CSI_SNAPSHOTTER_IMAGE="registry.k8s.io/sig-storage/csi-snapshotter:v6.2.2"
+export CSI_REGISTRAR_IMAGE="registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.8.0"
+
+export CK8S_VERSION="v1.30.0"                                                   # K8s Version of the cluster
+
+export VSPHERE_PROXY_DISABLE="#"                                                # Set to "#" to disable, or "" to enable the proxy configuration
```
neoaggelos, line 33: this is not `VSPHERE_` related.

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+export VSPHERE_FOLDER="vm"                                                      # The VM folder for your VMs. Set to "" to use the root vSphere folder
+export VSPHERE_TEMPLATE="ubuntu-1804-kube-v1.17.3"                              # The VM template to use for your management cluster.
+export CONTROL_PLANE_ENDPOINT_IP="192.168.9.230"                                # the IP that kube-vip is going to use as a control plane endpoint
+export VIP_NETWORK_INTERFACE="ens192"                                           # The interface that kube-vip should apply the IP to. Omit to tell kube-vip to autodetect the interface.
+export VSPHERE_TLS_THUMBPRINT="..."                                             # sha1 thumbprint of the vcenter certificate: openssl x509 -sha1 -fingerprint -in ca.crt -noout
+export EXP_CLUSTER_RESOURCE_SET="true"                                          # This enables the ClusterResourceSet feature that we are using to deploy CSI
+export VSPHERE_SSH_AUTHORIZED_KEY="ssh-rsa AAAAB3N..."                          # The public ssh authorized key on all machines in this cluster.
```
neoaggelos, line 17: remember, this could be optional

neoaggelos (commented):
Looks great, did another pass

templates/vsphere/template-variables.rc
neoaggelos: this is to prevent issues where this config is not set properly, and instead of a failure users get an invalid cluster template

bschimke95 (commented):
LGTM, waiting for @neoaggelos final approval

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+export CSI_DRIVER_IMAGE="gcr.io/cloud-provider-vsphere/csi/release/driver:v3.3.0"
+export CSI_SYNCER_IMAGE="gcr.io/cloud-provider-vsphere/csi/release/syncer:v3.3.0"
+export CSI_LIVENESS_PROBE_IMAGE="registry.k8s.io/sig-storage/livenessprobe:v2.10.0"
+export CSI_ATTACHER_IMAGE="registry.k8s.io/sig-storage/csi-attacher:v4.3.0"
+export CSI_RESIZER_IMAGE="registry.k8s.io/sig-storage/csi-resizer:v1.8.0"
+export CSI_PROVISIONER_IMAGE="registry.k8s.io/sig-storage/csi-provisioner:v3.5.0"
+export CSI_SNAPSHOTTER_IMAGE="registry.k8s.io/sig-storage/csi-snapshotter:v6.2.2"
+export CSI_REGISTRAR_IMAGE="registry.k8s.io/sig-storage/csi-node-driver-registrar:v2.8.0"
+
+export KUBERNETES_VERSION=""                    # K8s Version of the cluster, e.g. "v1.30.0"
```
neoaggelos, line 31: move the kubernetes version at the top, followed by control plane and worker node counts. then, all our cluster templates can be similar.

the order should be version -> cluster size -> vsphere configs -> ck8s configs

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,36 @@
...
+
+export PROXY_DISABLE="#"                        # Set to "#" to disable, or "" to enable the proxy configuration
```
neoaggelos, line 33: we cannot hardcode the proxy to squid.internal. Perhas move all the configs in here, as an example

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1514 @@
...
+        - echo "::1         ipv6-localhost ipv6-loopback localhost6 localhost6.localdomain6"
+          >/etc/hosts
+        - echo "127.0.0.1   {{ ds.meta_data.hostname }} {{ local_hostname }} localhost
+          localhost.localdomain localhost4 localhost4.localdomain4" >>/etc/hosts
+        - |
+          if [ -n "${VSPHERE_SSH_AUTHORIZED_KEY}" ]; then
+            echo "${VSPHERE_SSH_AUTHORIZED_KEY}" >> /root/.ssh/authorized_keys
+            chmod 700 /root/.ssh
+            chmod 600 /root/.ssh/authorized_keys
+          fi
```
neoaggelos, line 301: double check that these are required. for the authorized keys, maybe a `files` entry would be the better option.

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1514 @@
...
+      - echo "::1         ipv6-localhost ipv6-loopback localhost6 localhost6.localdomain6"
+        >/etc/hosts
+      - echo "127.0.0.1   {{ ds.meta_data.hostname }} {{ local_hostname }} localhost
+        localhost.localdomain localhost4 localhost4.localdomain4" >>/etc/hosts
+      - |
+        if [ -n "${VSPHERE_SSH_AUTHORIZED_KEY}" ]; then
+          echo "${VSPHERE_SSH_AUTHORIZED_KEY}" >> /root/.ssh/authorized_keys
+          chmod 700 /root/.ssh
+          chmod 600 /root/.ssh/authorized_keys
+        fi
```
neoaggelos, line 271: same here

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1514 @@
...
+      - snap install k8s --classic --edge
```
neoaggelos, line 272: we should not be promoting this for canonical k8s. instead, either override the `install.sh` script, or wait for a stable version before merging the PR.

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1514 @@
...
+  name: ${CLUSTER_NAME}-vsphere-proxy-config
+stringData:
+  proxy-config: |
+    ${PROXY_DISABLE}[Service]
+    ${PROXY_DISABLE}Environment="HTTPS_PROXY=http://squid.internal:3128"
+    ${PROXY_DISABLE}Environment="HTTP_PROXY=http://squid.internal:3128"
+    ${PROXY_DISABLE}Environment="https_proxy=http://squid.internal:3128"
+    ${PROXY_DISABLE}Environment="http_proxy=http://squid.internal:3128"
+    ${PROXY_DISABLE}Environment="NO_PROXY=10.0.0.0/8,192.168.0.0/16,127.0.0.1,localhost,0.0.0.0"
+    ${PROXY_DISABLE}Environment="no_proxy=10.0.0.0/8,192.168.0.0/16,127.0.0.1,localhost,0.0.0.0"
```
neoaggelos, line 13: we should not hardcode this. Consider whether we could have an option like `PROXY_CONFIGURATION` with a value like

```
PROXY_CONFIGURATION="
http_proxy=...
https_proxy=...
no_proxy=....
"
```

Then use `[Service] EnvironmentFile` and point to that file instead.

neoaggelos: See https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Command%20lines for the `EnvironmentFile` directive

neoaggelos (changes requested):

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1514 @@
...
+                tolerations:
+                  - effect: NoSchedule
+                    operator: Exists
+                  - effect: NoExecute
+                    operator: Exists
+            updateStrategy: {}
+          status: {}
+        owner: root:root
+        path: /capi/manifests/kube-vip.yaml
+        permissions: "0644"
```
neoaggelos, line 255: move these files to a secret, then specify `contentFrom`, such that CK8sControlPlane resource remains tidy

templates/vsphere/cluster-template.yaml
```diff
@@ -0,0 +1,1514 @@
...
+      - content: 127.0.0.1 localhost kubernetes
+        owner: root:root
+        path: /etc/kube-vip.hosts
+        permissions: "0644"
```
neoaggelos, line 259: validate that this is required

templates/vsphere/cluster-template.yaml
eaudetcobello: Keys moved to files. Will add a note to verify the preRunCommands before merging.

templates/vsphere/cluster-template.yaml
eaudetcobello: Will verify when we can create multi-node clusters.

templates/vsphere/template-variables.rc
```diff
@@ -0,0 +1,55 @@
...
+export WORKER_MACHINE_CPU_COUNT=2
+export WORKER_MACHINE_MEMORY_MB=2048
+export WORKER_MACHINE_DISK_GB=25
+
+# Source: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/docs/getting_started.md
+
+export VSPHERE_USERNAME=""                      # The username used to access the remote vSphere endpoint
+export VSPHERE_PASSWORD=""                      # The password used to access the remote vSphere endpoint
+
+export VSPHERE_SERVER=""                        # The vCenter server IP or FQDN, e.g. "10.0.0.1"
```
bschimke95, line 32: for AWS you don't set defaults. Please either update the AWS env (in another PR) or this rc file
https://github.com/canonical/cluster-api-k8s/blob/main/templates/aws/template-variables.rc

eaudetcobello: @bschimke95 Not sure what you mean here

bschimke95: You don't set an empty string in the AWS template e.g. `export AWS_PUBLIC_IP=`

bschimke95 (approved):
LGTM

PR #26, opened 2024-07-11
Ignore kubernetes version prefix for machine filter

bschimke95:
This has caused a broken cluster now for several times. Let's make sure we ignore the prefix in the comparison.

### DIFF
--- a/pkg/machinefilters/machine_filters.go
+++ b/pkg/machinefilters/machine_filters.go
@@ -18,6 +18,7 @@ package machinefilters
 
 import (
 	"reflect"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


pkg/machinefilters/machine_filters.go
```diff
@@ -78,7 +79,7 @@ func MatchesKubernetesVersion(kubernetesVersion string) Func {
 		if machine.Spec.Version == nil {
 			return false
 		}
-		return *machine.Spec.Version == kubernetesVersion
+		return strings.TrimPrefix(*machine.Spec.Version, "v") == strings.TrimPrefix(kubernetesVersion, "v")
```
neoaggelos, line 82: this is OK to start with, but will need more iterating when we are doing the in-place upgrades work.

neoaggelos (approved):

PR #29, opened 2024-07-24
Add rollout upgrade integration test

bschimke95:
# Summary

Adds a test that creates a HA workload cluster and upgrades it from 1.29 Kubernetes to 1.30.

# Implementation details

* Non-HA clusters require in-place updates which are currently not supported
* We don't have snapd on the docker infra and cannot install different versions via `snap install`. That's why we build two different docker images and replace the infrastructure reference accordingly along with the upgrade. On other infra providers we would simply change the version and not the image.
* Right now, we inject a Kubernetes override into the docker build process as we don't have another minor release besides 1.30. Once 1.31 is out, we can bump the test versions, upgrade from 1.30 -> 1.31 and remove this hack in the build process.

# Notes
This tests require https://github.com/canonical/microcluster/pull/197 to land in microcluster and https://github.com/canonical/k8s-snap/pull/560 and https://github.com/canonical/k8s-snap/pull/562 to land in k8s-snap.


test/e2e/cluster_upgrade.go
```diff
@@ -157,6 +157,7 @@ func ClusterUpgradeSpec(ctx context.Context, inputGetter func() ClusterUpgradeSp
 			ClusterProxy:                input.BootstrapClusterProxy,
 			Cluster:                     result.Cluster,
 			ControlPlane:                result.ControlPlane,
+			UpgradeMachineTemplate:      ptr.To(fmt.Sprintf("%s-control-plane-1.30", clusterName)),
```
addyess, line 160: Does it feel odd to hard-code these version numbers into the test?  I suppose it's fine b/c it could be any version right?  like 1.xx

addyess (commented):
Just comments.  I don't feel like I can give better feedback

test/e2e/cluster_upgrade.go
bschimke95: Yes, we don't test a particular version but rather if the upgrade mechanism of the provider is working as expected.

pkg/ck8s/workload_cluster.go
```diff
@@ -281,8 +279,8 @@ func (w *Workload) doK8sdRequest(ctx context.Context, method, endpoint string, r
 	if responseBody.Error != "" {
 		return fmt.Errorf("k8sd request failed: %s", responseBody.Error)
 	}
-	if responseBody.Metadata == nil {
-		// Nothing to decode
+	if responseBody.Metadata == nil || response == nil {
```
bschimke95, line 282: drive-by fix: This would not cause issues but pollutes the logs with an error.

test/e2e/cluster_upgrade_test.go
```diff
@@ -25,12 +25,11 @@ import (
...
 
 var _ = Describe("Workload cluster upgrade [CK8s-Upgrade]", func() {
-	BeforeEach(func() {
-		// TODO(bschimke): Remove once we find a way to run e2e tests with other infrastructure providers that support snap.
-		Skip("Skipping the upgrade tests as snap does not work on CAPD.")
-	})
-
 	Context("Upgrading a cluster with 1 control plane", func() {
+		It("Non-HA upgrades require in-place upgrades which are not supported yet.",
+			// TODO(ben): Enable this test once we have support for in-place upgrades.
```
bschimke95, line 30: @berkayoz This is the test we need to enable again once the in-place upgrades are there.

templates/aws/ccm.yaml
neoaggelos, line 1: this in unrelated for this PR

templates/docker/Dockerfile
```diff
@@ -36,15 +36,21 @@ FROM $BUILD_BASE AS builder
 ARG REPO=https://github.com/canonical/k8s-snap
 ARG BRANCH=main
 
+## Override the Kubernetes version from the branch.
+## Note(ben): We only have 1.30-release branches for k8s-snap right now.
+## For the rollout upgrades, we need to have a different minor version.
+## This is a temporary solution until we have a 1.31 release branch.
+ARG KUBERNETES_VERSION_OVERRIDE=""
```
neoaggelos, line 43: `ARG KUBERNETES_VERSION` should suffice, and no need for the extra comments. It's OK to have a build argument to specify the exact version if we want

templates/docker/Dockerfile
```diff
@@ -86,6 +92,12 @@ RUN /src/k8s-snap/build-scripts/build-component.sh helm
 
 ## kubernetes build
 FROM builder AS build-kubernetes
+ENV KUBERNETES_VERSION_OVERRIDE=${KUBERNETES_VERSION_OVERRIDE}
+RUN if [ -n "$KUBERNETES_VERSION_OVERRIDE" ]; then \
+      echo "Overwriting Kubernetes version with $KUBERNETES_VERSION_OVERRIDE"; \
+      echo "$KUBERNETES_VERSION_OVERRIDE" > /src/k8s-snap/build-scripts/components/kubernetes/version; \
+      cat /src/k8s-snap/build-scripts/components/kubernetes/version; \
```
neoaggelos, line 99: this can be just:
```
[ -n "$KUBERNETES_VERSION" ] && echo "$KUBERNETES_VERSION" > /src/k8s-snap/build-scripts/components/kubernetes/version
```

test/e2e/data/infrastructure-docker/cluster-template-upgrades.yaml
```diff
@@ -0,0 +1,136 @@
...
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: DockerMachineTemplate
+metadata:
+  name: ${CLUSTER_NAME}-control-plane-1.30
+  namespace: ${NAMESPACE}
+spec:
+  template:
+    spec:
+      customImage: k8s-snap:dev-1.30
```
neoaggelos, line 73: why do we deploy a second machine template? we should simply apply the updated version of the `DockerMachineTemplate` resource

test/e2e/helpers.go
```diff
@@ -577,6 +578,17 @@ func UpgradeControlPlaneAndWaitForUpgrade(ctx context.Context, input UpgradeCont
...
+	newInfrastructureRef := corev1.ObjectReference{
+		APIVersion: "infrastructure.cluster.x-k8s.io/v1beta1", // Adjust based on your infrastructure API version
+		Kind:       "DockerMachineTemplate",
+		Name:       fmt.Sprintf("%s-control-plane-1.30", input.Cluster.Name),
+		Namespace:  input.ControlPlane.Spec.MachineTemplate.InfrastructureRef.Namespace,
+	}
+
+	// Update the infrastructureRef
+	input.ControlPlane.Spec.MachineTemplate.InfrastructureRef = newInfrastructureRef
+
```
neoaggelos, line 591: we should instead just update the machine template with the new image

test/e2e/config/ck8s-docker.yaml
```diff
@@ -84,8 +85,8 @@ providers:
 
 variables:
   KUBERNETES_VERSION_MANAGEMENT: "v1.28.0"
-  KUBERNETES_VERSION: "v1.30.0"
-  KUBERNETES_VERSION_UPGRADE_TO: "v1.30.1"
+  KUBERNETES_VERSION: "v1.29.6"
+  KUBERNETES_VERSION_UPGRADE_TO: "v1.30.2"
```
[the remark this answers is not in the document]
bschimke95: This actually does not make any difference.
We only compare for the minor version and build the relevant images in the e2e.yaml Github workflow rather then using this value here.

.github/workflows/e2e.yaml
```diff
@@ -11,6 +11,14 @@ jobs:
...
+        name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          # We run into rate limiting issues if we don't authenticate
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Check out repo
         uses: actions/checkout@v4
       - name: Install requirements
```
neoaggelos, line 24: Just a note: make sure to avoid too high coupling between the workflow itself and the setting up of the CI (e.g. runners). It works now, but the danger would be that it becomes very brittle and one of those things that people are afraid to touch.

.github/workflows/e2e.yaml
```diff
@@ -25,26 +33,30 @@ jobs:
       - name: Build k8s-snap image
         run: |
           cd templates/docker
-          sudo docker build . -t k8s-snap:dev
+          sudo docker build . -t k8s-snap:dev-old --build-arg BRANCH=main --build-arg KUBERNETES_VERSION=v1.29.6
+          sudo docker build . -t k8s-snap:dev-new --build-arg BRANCH=main
```
neoaggelos, line 37: should this be in a wrapper script? where you build all necessary images? this version will need to be updated eventually, it would rather not be a hardcoded value on a github action

test/e2e/config/ck8s-docker.yaml
```diff
@@ -84,8 +85,8 @@ providers:
 
 variables:
   KUBERNETES_VERSION_MANAGEMENT: "v1.28.0"
-  KUBERNETES_VERSION: "v1.30.0"
-  KUBERNETES_VERSION_UPGRADE_TO: "v1.30.1"
+  KUBERNETES_VERSION: "v1.29.6"
+  KUBERNETES_VERSION_UPGRADE_TO: "v1.30.3"
   IP_FAMILY: "IPv4"
```
neoaggelos, line 90: especially when we already repeat the configs here. i can already see confusion about "i updated this value, why does it not work as i expect?"

neoaggelos (commented):
Overall LGTM, left some comments around structure

.github/workflows/e2e.yaml
bschimke95: You mean the Github login should not be a separate step? Where would you put this instead?

test/e2e/config/ck8s-docker.yaml
```diff
@@ -51,9 +51,10 @@ providers:
           - old: "imagePullPolicy: Always"
             new: "imagePullPolicy: IfNotPresent"
     files:
-      - sourcePath: "../data/infrastructure-docker/cluster-template.yaml"
       - sourcePath: "../data/infrastructure-docker/cluster-template-kcp-remediation.yaml"
       - sourcePath: "../data/infrastructure-docker/cluster-template-md-remediation.yaml"
+      - sourcePath: "../data/infrastructure-docker/cluster-template-upgrades.yaml"
+      - sourcePath: "../data/infrastructure-docker/cluster-template.yaml"
```
HomayoonAlimohammadi, line 57: Just being curious, does the ordering matter here?

test/e2e/data/infrastructure-docker/cluster-template-md-remediation.yaml
```diff
@@ -57,7 +57,7 @@ metadata:
 spec:
   template:
     spec:
-      customImage: k8s-snap:dev
+      customImage: k8s-snap:dev-old
```
HomayoonAlimohammadi, line 60: again, just being curious, is there any particular reason that we used `dev-old` here and `dev-new` in the other non-upgrade templates (e.g. `cluster-template-kcp-remediation.yaml`)?

test/e2e/data/infrastructure-docker/cluster-template-upgrades.yaml
```diff
@@ -0,0 +1,135 @@
...
+      kind: DockerMachineTemplate
+      name: ${CLUSTER_NAME}-control-plane-old
+  spec:
+    airGapped: true
+    controlPlane:
+      extraKubeAPIServerArgs:
+        --anonymous-auth: "true"
+  replicas: ${CONTROL_PLANE_MACHINE_COUNT}
+  version: ${KUBERNETES_VERSION}
+# Initial template for the machine deployment
```
HomayoonAlimohammadi, line 51: maybe we can move this to the next line (after `---`)? I think it's more readable that way.

test/e2e/data/infrastructure-docker/cluster-template-upgrades.yaml
```diff
@@ -0,0 +1,135 @@
...
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: DockerMachineTemplate
+metadata:
+  name: ${CLUSTER_NAME}-control-plane-old
+  namespace: ${NAMESPACE}
+spec:
+  template:
+    spec:
+      customImage: k8s-snap:dev-old
+# After upgrade template for the machine deployment
```
HomayoonAlimohammadi, line 62: this as well

test/e2e/data/infrastructure-docker/cluster-template-upgrades.yaml
```diff
@@ -0,0 +1,135 @@
...
+        configRef:
+          apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
+          kind: CK8sConfigTemplate
+          name: ${CLUSTER_NAME}-md-0
+      infrastructureRef:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+        kind: DockerMachineTemplate
+        name: ${CLUSTER_NAME}-md-old-0
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
HomayoonAlimohammadi, line 107: do we need to add any comments for worker nodes as well?

test/e2e/helpers.go
```diff
@@ -607,14 +661,17 @@ func WaitForNodesReady(ctx context.Context, input WaitForNodesReadyInput) {
 		}
 		nodeReadyCount := 0
 		for _, node := range nodeList.Items {
+			fmt.Fprintf(GinkgoWriter, "versions: %s %s\n", semver.MajorMinor(node.Status.NodeInfo.KubeletVersion), semver.MajorMinor(input.KubernetesVersion))
```
HomayoonAlimohammadi, line 664: maybe we can add more context to this line, something like:
```suggestion
			fmt.Fprintf(GinkgoWriter, "KubeletVersion: %s, KubernetesVersion: %s\n", semver.MajorMinor(node.Status.NodeInfo.KubeletVersion), semver.MajorMinor(input.KubernetesVersion))
```

HomayoonAlimohammadi (approved):
LGTM, Thanks a lot @bschimke95! Just some minor comments and questions.

test/e2e/config/ck8s-docker.yaml
bschimke95: I don't think so, these are just the specifications of template files.

test/e2e/data/infrastructure-docker/cluster-template-md-remediation.yaml
bschimke95: Not really, let's use `dev-old` everywhere.

PR #30, opened 2024-07-29
Add in-place upgrade proposal

berkayoz:
Add in-place upgrade proposal

### DIFF
--- a/docs/proposals/001-in-place-upgrades.md
+++ b/docs/proposals/001-in-place-upgrades.md
@@ -58,14 +58,8 @@ clear example of "before" and "after".
 
 Users will be able to perform in-place upgrades per machine basis by running:
 ```sh
-kubectl annotate machine <machine-name> k8sd.io/in-place-upgrade-to={upgrade-option}
+kubectl annotate machine <machine-name> v1beta2.k8sd.io/in-place-upgrade-to={upgrade-option}
 ```
-
-Users can also perform in-place upgrades on the entire cluster by running:
-```sh
-kubectl annotate cluster <cluster-name> k8sd.io/in-place-upgrade-to={upgrade-option}
-```
-This would upgrade machines belonging to `<cluster-name>` one by one.
 
 `{upgrade-option}` can be one of:
 * `channel=<channel>` which would refresh the machine to the provided channel e.g. `channel=1.31-classic/stable`
@@ -101,19 +95,25 @@ below, or serve as reference for future proposals.
 ### Cluster-wide Orchestration
 A cluster controller called `ClusterReconciler` is added which would perform the one-by-one in-place upgrade of the entire workload cluster. 
 
-The controller would propagate the `k8sd.io/in-place-upgrade-to` annotation on the `Cluster` object by adding this annotation one-by-one to all the machines that is owned by this cluster. 
+Users perform in-place upgrades on the entire cluster by running:
+```sh
+kubectl annotate cluster <cluster-name> v1beta2.k8sd.io/in-place-upgrade-to={upgrade-option}
+```
+This would upgrade machines belonging to `<cluster-name>` one by one.
+
+The controller would propagate the `v1beta2.k8sd.io/in-place-upgrade-to` annotation on the `Cluster` object by adding this annotation one-by-one to all the machines that is owned by this cluster. 
 
 The reconciler would perform upgrades in 2 separate phases for control-plane and worker machines. 
 
 A Kubernetes API call listing the objects of type `Machine` and filtering with `ownerRef` would produce the list of machines owned by the cluster. For each phase controller would iterate over this list filtering by the machine type, annotating the machines and waiting for the operation to complete on each iteration.
 
-The reconciler should ignore a machine if `k8sd.io/in-place-upgrade-status` is already set to `in-progress`. 
+The reconciler should not trigger the upgrade endpoint if `v1beta2.k8sd.io/in-place-upgrade-status` is already set to `in-progress` on the machine.
 
 Once upgrades of the underlying machines are finished:
-* `k8sd.io/in-place-upgrade-to` annotation on the `Cluster` would be removed
-* `k8sd.io/in-place-upgrade-release` annotation on the `Cluster` would be added/updated with the used `{upgrade-option}`.
+* `v1beta2.k8sd.io/in-place-upgrade-to` annotation on the `Cluster` would be removed
+* `v1beta2.k8sd.io/in-place-upgrade-release` annotation on the `Cluster` would be added/updated with the used `{upgrade-option}`.
 
-This will be introduced and explained more extensively in another proposal.
+This process can be adapted to use `CK8sControlPlane` and `MachineDeployment` objects instead to be able to upgrade control-plane and worker nodes separately. This will be introduced and explained more extensively in another proposal.
 
 ### Upgrades of Underlying OS and Dependencies
 The in-place upgrades only address the upgrades of Canonical Kubernetes and it's respective dependencies. Which means changes on the OS front/image would not be handled since the underlying machine image stays the same. This would be handled by a rolling upgrade as usual.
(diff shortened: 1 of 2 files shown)


docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+A machine controller called `MachineReconciler` is added which would perform the in-place upgrade if `k8sd.io/in-place-upgrade` annotation is set on the machine.
+
+The controller would use the value of this annotation to make an endpoint call to the `/x/capi/snap-refresh` through `k8sd-proxy`. 
+
+The result of this operation will be communicated back to the user via the `k8sd.io/in-place-upgrade-status` annotation. Values being:
+
+* `done` for a successful upgrade
+* `failed` for a failed upgrade
+
+A failed upgrade could be re-triggered by removing the `k8sd.io/in-place-upgrade` annotation and re-adding it to the machine.
```
louiseschmidtgen, line 145: Should we be re-trying the upgrade until the annotation is removed (upgraded succesfully). So not just on the annotation being set?

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+The `ck8sconfig_controller` should check for the `k8sd.io/in-place-upgrade` annotation both on the `Machine` and on the owner `Cluster` object. The value of one of these annotations should be used instead of the `version` field while generating a cloud-init script for a new machine. The annotation on the `Machine` object should take precedence.
```
louiseschmidtgen, line 147:
```suggestion
The `ck8s_config_controller` should check for the `k8sd.io/in-place-upgrade` annotation both on the `Machine` and on the owner `Cluster` object. The value of one of these annotations should be used instead of the `version` field while generating a cloud-init script for a new machine. The annotation on the `Machine` object should take precedence.
```

docs/proposals/001-in-place-upgrades.md
berkayoz: We could certainly retry the upgrades, even then I believe there should be a maximum number of retries. The `k8sd.io/in-place-upgrade` annotation won't be removed after an upgrade, we add the `k8sd.io/in-place-upgrade-status` annotation to report the result and mark the operation as finished. 

My worry about this is that failed retries could disrupt the workload cluster somehow. It feels like a failed upgrade is less likely to be resolved by just retrying and it will possibly require administrative/user attention.

docs/proposals/001-in-place-upgrades.md
berkayoz: This was a direct reference to the actual controller file in the repo while I was drafting, we can change this to something generic like `CK8sConfigReconciler`

louiseschmidtgen, line 147: I wonder if by default an update on the cluster annotation should update all nodes. Each machine would always have the version as an annotation or something like `no-update`.

Then we could retry updating the machine version to the cluster-version until they are the same?

I think this could be useful in a sense that the control over the update is closer to the management cluster level rather than the individual level.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+This would prevent adding nodes with an outdated version and possibly breaking the cluster due to a version mismatch.
+
+## ControlPlane Provider Changes
+<!--
+This section MUST mention any changes to the controlplane provider.
+-->
+A cluster controller called `ClusterReconciler` is added which would perform the one-by-one in-place upgrade of the entire workload cluster. 
+
+The controller would propagate the `k8sd.io/in-place-upgrade` annotation on the `Cluster` object by adding this annotation one-by-one to all the machines that is owned by this cluster.
```
bschimke95, line 156: I would add:
* a machine that already contains this annotation will not be overwritten
* after the upgrade is done, all annotations will be removed, but *only* if the annotation matches the initial cluster annotation. Otherwise this would remove any machine level annotation.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+## Documentation Changes
+<!--
+This section MUST mention any new documentation that is required for the new
+feature. Most features are expected to come with at least a How-To and an
+Explanation page.
+
+In this section, it is useful to think about any existing pages that need to be
+updated (e.g. command outputs).
+-->
+`How-To` page on performing in-place upgrades should be created.
```
bschimke95, line 177: We definitely need to explain somewhere why we are doing this custom annotation dance (-> CAPI does not yet support different upgrade strategies).

Not sure, if an extra page would make sense (where we could also explain the difference between in-place/rollout upgrades), or if we should just add it as a note to the how-to.

Probably up to @evilnick to decide.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+
+We should be aware of the [metadata propagation](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/metadata-propagation) performed by the upstream controllers. Some meetadata is propagated in-place, which can ultimately propagate all the way down to the `Machine` objects. This could potentially flood the cluster with upgrades if machines get annotated at the same time. The cluster wide upgrade is handled through the annotation on the actual Cluster object due to this reason.
+
+Updating the `version` field would trigger rolling updates by default, with the only difference than upstream being the precedence of the version value provided in the annotations. 
+
+The in-place upgrades only address the upgrades of Canonical Kubernetes and it's respective dependencies. Which means changes on the OS front/image would not be handled since the underlying machine image stays the same. This would be handled by a rolling upgrade as usual.
```
bschimke95, line 221: This would probably fit better in the `out-of-scope` section

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+particular commits, not branches, such that any links remain valid going
+forward).
+
+This is useful as it allows the proposal owner to not be the person that
+implements it.
+-->
+
+The annotation method is chosen due to the "immutable infrastructure" assumption CAPI currently has. Which means updates are always done by creating new machines and fields are immutable. This might also pose some challenges on displaying accurate Kubernetes version information through CAPI.
+
+We should be aware of the [metadata propagation](https://cluster-api.sigs.k8s.io/developer/architecture/controllers/metadata-propagation) performed by the upstream controllers. Some meetadata is propagated in-place, which can ultimately propagate all the way down to the `Machine` objects. This could potentially flood the cluster with upgrades if machines get annotated at the same time. The cluster wide upgrade is handled through the annotation on the actual Cluster object due to this reason.
```
bschimke95, line 217: typo: `metadata` instead of `meetadata`

bschimke95 (commented):
Great proposal @berkayoz 
Did a first pass

docs/proposals/001-in-place-upgrades.md
bschimke95: Let's not do it. Separation of concerns - you don't want to have upgrade logic spread over multiple controllers.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+## API Changes
+<!--
+This section MUST mention any changes to the k8sd API, or any additional API
+endpoints (and messages) that are required for this proposal.
+
+Unless there is a particularly strong reason, it is preferable to add new v2/v3
+APIs endpoints instead of breaking the existing APIs, such that API clients are
+not affected.
+-->
+### `POST /x/capi/snap-refresh`
```
bschimke95, line 114: The `snap-refresh` endpoint is not really CAPI specific. Right now, we will only use it there but who knows...

Maybe `/x/snap/refresh` makes more sense?

docs/proposals/001-in-place-upgrades.md
berkayoz: Agreeing with @bschimke95 here, the upgrades are scoped to `Machine`'s. The `Cluster` level annotation is simply a helper to offload the work of annotating all cluster machines manually.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,221 @@
...
+<!--
+This section MUST mention any changes to the bootstrap provider.
+-->
+
+A machine controller called `MachineReconciler` is added which would perform the in-place upgrade if `k8sd.io/in-place-upgrade` annotation is set on the machine.
+
+The controller would use the value of this annotation to make an endpoint call to the `/x/capi/snap-refresh` through `k8sd-proxy`. 
+
+The result of this operation will be communicated back to the user via the `k8sd.io/in-place-upgrade-status` annotation. Values being:
+
```
bschimke95, line 141: I would add `in-progress`

docs/proposals/001-in-place-upgrades.md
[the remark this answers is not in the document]
berkayoz: This is about how to possibly handle rolling upgrades(a new machine is created) when the cluster version is ahead through in-place upgrades.

docs/proposals/001-in-place-upgrades.md
berkayoz: Updated the annotations, should be mostly addressed on how machines are picked up / ignored.

docs/proposals/001-in-place-upgrades.md
berkayoz: I am a +1 on this as well, I can update the proposal depending on what kind of change is expected(new page, a section on a page etc.)

docs/proposals/001-in-place-upgrades.md
berkayoz: Wouldn't we need to create a CAPI specific endpoint like we've done for `x/capi/remove-node` to do CAPI specific auth?

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+-->
+
+Users will be able to perform in-place upgrades per machine basis by running:
+```sh
+kubectl annotate machine <machine-name> k8sd.io/in-place-upgrade-to={upgrade-option}
+```
+
+Users can also perform in-place upgrades on the entire cluster by running:
+```sh
+kubectl annotate cluster <cluster-name> k8sd.io/in-place-upgrade-to={upgrade-option}
```
ktsakalozos, line 66: You want to upgrade the control-plane and the workers in separate steps.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+* `k8sd.io/in-place-upgrade-current` annotation on the `Machine` would be added/updated with the used `{upgrade-option}`.
+* `k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `done`
+
+After a failed upgrade:
+* `k8sd.io/in-place-upgrade-failure` annotation on the `Machine` would be added/updated with the failure message
+* `k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `failed`
+
+The reconciler should ignore the upgrade if `k8sd.io/in-place-upgrade-status` is already set to `in-progress` on the machine. 
+
+#### Changes for Rolling Upgrades and Creating New Machines
```
ktsakalozos, line 162: I assume this covers the case of scaling up a cluster.

berkayoz: Indeed this does cover scaling as well, I'll add add this as well thanks.

docs/proposals/001-in-place-upgrades.md
berkayoz: Good point, I'll update it with more details on this.

docs/proposals/001-in-place-upgrades.md
bschimke95: true, good point. Ignore what I wrote.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+Which means the value from the annotation on the `Machine` would be used first if found.
+
+Using an annotation value requires changing the `install.sh` file to perform the relevant snap operation based on the option.
+* `snap install k8s --classic --channel <channel>` for `Channel`
+* `snap install k8s --classic --revision <revision>` for `Revision`
+* `snap install <path-to-snap> --classic --dangerous --name k8s` for `LocalPath`
+
+When a rolling upgrade is triggered the `LocalPath` option requires the newly created machine to contain the local `*.snap` file. This usually means the machine image used by the infrastructure provider should be updated to contain this image. This file could possibly be sideloaded in the cloud-init script before installation.
+
+This operation should not be performed if `install.sh` script is overridden by the user in the manifests.
```
bschimke95, line 179: How do we detect that?

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+* `in-progress` for an upgrade currently in progress
+* `done` for a successful upgrade
+* `failed` for a failed upgrade
+
+After an upgrade process begins:
+* `k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `in-progress`
+
+After a successfull upgrade:
+* `k8sd.io/in-place-upgrade-to` annotation on the `Machine` would be removed
+* `k8sd.io/in-place-upgrade-current` annotation on the `Machine` would be added/updated with the used `{upgrade-option}`.
```
bschimke95, line 153: nit: let's use `release` or `variant`. `Current` IMHO is not that obvious

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+* `k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `done`
+
+After a failed upgrade:
+* `k8sd.io/in-place-upgrade-failure` annotation on the `Machine` would be added/updated with the failure message
```
bschimke95, line 157: let's make this an `upgrade-status-message`. We might want to have human-friendly messages for other statuses as well (e.g. `in-progress` could have something like `upgrade from ... to ... in progress`)

docs/proposals/001-in-place-upgrades.md
berkayoz: This will be done my making sure we don't override the `install.sh` script with the in-place options if there is a user provided file for it. We might not need to do anything specific other than making sure the `ExtraFiles` take precedence.

docs/proposals/001-in-place-upgrades.md
berkayoz: @bschimke95 My approach for this would be to check which machines belong to the control plane, perform upgrades for those first then proceed with workers. We could wait for the upgrades on control plane machines to be completed before starting the worker node process. A possible `k8sd.io/in-place-upgrade-phase` with `control-plane` and `workers` annotation could control this. The cluster would get annotated with `k8sd.io/in-place-upgrade-phase` with `workers` once `control-plane` phase finishes.

bschimke95: Sounds good. In your suggestion, would the user set `k8sd.io/in-place-upgrade-phase` or would this be set by the provider?

I read @ktsakalozos comment as "We need a way to let the user control this", right?

berkayoz: This would be set by the provider and would be removed from the `Cluster` object once the upgrades are completed. 

For a more user controlled upgrade approach I feel it might be better/possible to go through `CK8sControlPlane` and `MachineDeployment` objects. It would be split more logically and users could annotate those in order as desired.

bschimke95: Good point. How would this look like?
Cluster -> annotates CK8sControlPlane and MachineDeployment -> annotate respective machines? So annotating `Cluster` would basically be a convenience layer on the convenience layer.

docs/proposals/001-in-place-upgrades.md
berkayoz: Having a status message in annotations does not feel right to me, maybe we could use events for this so users could see through `kubectl describe`.

docs/proposals/001-in-place-upgrades.md
berkayoz: Correct, we would annotate child objects of the `Cluster`. This seems more balanced in terms of convenience and user control.

docs/proposals/001-in-place-upgrades.md
bschimke95: Is `in-place-upgrade-failure` not a status message - with the limitation that it only is populated on failures?

berkayoz: Oh it is for sure, I was wondering if we should move this out of the annotations completely. Sorry seems like I worded it a bit weird 🙂

docs/proposals/001-in-place-upgrades.md
neoaggelos: I would move this to an "Out Of Scope" section, left as a future consideration for now. Just so that the focus can be on getting in-place upgrades to work properly, be well tested.

Then, we can return to build a nice orchestration around this.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+	// LocalPath is the local path to use to refresh the snap.
+	LocalPath string
+}
+```
+
+`POST /x/capi/snap-refresh` performs the in-place upgrade with the given options.
+
+The upgrade can be either done with a `Channel`, `Revision` or a local `*.snap` file provided via `LocalPath`. The value of `LocalPath` should be an absolute path.
+
+This endpoint should use `ValidateCAPIAuthTokenAccessHandler("capi-auth-token")` for authentication.
```
neoaggelos, line 131: Be aware that this token should be different from the main token we use to authenticate in general. Otherwise, any worker nodes will be able to know this token (and possibly escalate their access through k8sd).

Should we have a `capiJoinToken` and a `capiRefreshToken`? Maybe even add a separate token _per node_ on the token secret (using a `refresh-token::$nodename` key). that way, if someone traces this token, they cannot authenticate on other k8sd nodes.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+## Bootstrap Provider Changes
+<!--
+This section MUST mention any changes to the bootstrap provider.
+-->
+
+A machine controller called `MachineReconciler` is added which would perform the in-place upgrade if `k8sd.io/in-place-upgrade-to` annotation is set on the machine.
+
+The controller would use the value of this annotation to make an endpoint call to the `/x/capi/snap-refresh` through `k8sd-proxy`. 
+
+The result of this operation will be communicated back to the user via the `k8sd.io/in-place-upgrade-status` annotation. Values being:
```
neoaggelos, line 142: +1 to using annotations for tracking this, this is also a lot easier to deal with in the code to check what the current status is.

In order to then communicate it to users (and make sure that the cluster status reflects any possible errors, we should consider adding a condition using a well-known name) that we set according to the result of the upgrade.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+After a successfull upgrade:
+* `k8sd.io/in-place-upgrade-to` annotation on the `Machine` would be removed
+* `k8sd.io/in-place-upgrade-current` annotation on the `Machine` would be added/updated with the used `{upgrade-option}`.
+* `k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `done`
+
+After a failed upgrade:
+* `k8sd.io/in-place-upgrade-failure` annotation on the `Machine` would be added/updated with the failure message
+* `k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `failed`
+
+The reconciler should ignore the upgrade if `k8sd.io/in-place-upgrade-status` is already set to `in-progress` on the machine.
```
neoaggelos, line 160: Also, it would probably be wise to have a configurable (?) parallelism for multiple machines at once. With rollout replace, machines are rotated one by one, we might want to mimic the same behaviour here (== if any other in-place upgrade is already in progress, just requeue the request).

docs/proposals/001-in-place-upgrades.md
neoaggelos: We fully control the generated cloud-init here. So we could just replace the default install script with one that does whatever is needed based on the annotation (instead of using the embedded `scriptInstall`).

Then, if users have specified a custom `install.sh` script, the implementation already covers this case for us (it will come later in `extraFiles` of cloud-init, so it will override the default one.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+
+The controller would propagate the `k8sd.io/in-place-upgrade-to` annotation on the `Cluster` object by adding this annotation one-by-one to all the machines that is owned by this cluster. 
+
+A Kubernetes API call listing the objects of type `Machine` and filtering with `ownerRef` would produce the list of machines owned by the cluster. The controller then would iterate over this list, annotating machines and waiting for the operation to complete on each iteration.
+
+The reconciler should ignore a machine if `k8sd.io/in-place-upgrade-status` is already set to `in-progress`. 
+
+Once upgrades of the underlying machines are finished:
+* `k8sd.io/in-place-upgrade-to` annotation on the `Cluster` would be removed
+* `k8sd.io/in-place-upgrade-current` annotation on the `Cluster` would be added/updated with the used `{upgrade-option}`.
```
neoaggelos, line 197: This is a good approach. I'd recommend moving this as "Out Of Scope" to start with, until we have some confidence on the feature itself. It would be easy to lose cycles iterating on this orchestration layer, rather than building and testing out the in place upgrades.

"Out of scope" means: keep the text, _but_ do not commit to an implementation decision just yet.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,257 @@
...
+<!--
+This section MUST describe any user-facing changes that this feature brings, if
+any. If an API change is required, the affected endpoints MUST be mentioned. If
+the output of any k8s command changes, the difference MUST be mentioned, with a
+clear example of "before" and "after".
+-->
+
+Users will be able to perform in-place upgrades per machine basis by running:
+```sh
+kubectl annotate machine <machine-name> k8sd.io/in-place-upgrade-to={upgrade-option}
```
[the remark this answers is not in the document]
berkayoz: I'd say Machines feel better in UX terms.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,308 @@
...
+* An `InPlaceUpgradeDone` event is added to the `Machine` with the `Successfully performed in place upgrade with {upgrade-option}` message.
+
+After a failed upgrade:
+* `v1beta2.k8sd.io/in-place-upgrade-status` annotation on the `Machine` would be added/updated with `failed`
+* `v1beta2.k8sd.io/in-place-change-id` annotation on the `Machine` would be removed
+* An `InPlaceUpgradeFailed` event is added to the `Machine` with the `Failed to perform in place upgrade with option {upgrade-option}: {error}` message.
+
+A custom condition with type `InPlaceUpgradeStatus` can also be added to relay these information. 
+
+The reconciler should not trigger the upgrade endpoint if `v1beta2.k8sd.io/in-place-upgrade-status` is already set to `in-progress` on the machine. Instead
```
HomayoonAlimohammadi, line 227: I think the follow up on "Instead" is missing.

docs/proposals/001-in-place-upgrades.md
```diff
@@ -0,0 +1,308 @@
...
+In this section, you SHOULD go into detail about how the proposal can be
+implemented. If needed, link to specific parts of the code (link against
+particular commits, not branches, such that any links remain valid going
+forward).
+
+This is useful as it allows the proposal owner to not be the person that
+implements it.
+-->
+
+The annotation method is chosen due to the "immutable infrastructure" assumption CAPI currently has. Which means updates are always done by creating new machines and fields are immutable. This might also pose some challenges on displaying accurate Kubernetes version information through CAPI.
```
HomayoonAlimohammadi, line 304: I think now the `version` column in `kubectl get machines` is not accurate, right? do we have any solutions for that? or is it acceptable?

HomayoonAlimohammadi (commented):
Thanks a lot @berkayoz! Left some minor comments.

docs/proposals/001-in-place-upgrades.md
berkayoz: That is correct, sadly the `version` field is used in upstream controllers for reconciling machine sets. Setting that field has a lot of complexity, I believe it is acceptable to have inaccurate information there. IIRC the `kubectl describe machine` lists component versions for proxy, kubelet etc. which shows a correct value.

HomayoonAlimohammadi: got it, thanks a lot berkay!

bschimke95 (approved):
LGTM, great work

HomayoonAlimohammadi:
follow up: [cluster-wide in-place upgrade PR](https://github.com/canonical/cluster-api-k8s/pull/55)

PR #32, opened 2024-07-31
disable k8s node cleanup and let CAPI handle it

bschimke95:
This pull request addresses an issue where nodes are not properly drained in Cluster API before upgrades. The current CK8s remove hook removes the node from the Kubernetes cluster, which conflicts with the CAPI core controllers responsible for node removal within the CAPI context. As a result, when our controllers remove the node, the core controller loses track and only removes the resource reference in the management cluster.

This PR disables the Kubernetes node removal in the k8s-snap.

Requires https://github.com/canonical/k8s-snap/pull/577 to be merged. See there for more details.

### DIFF
--- a/pkg/ck8s/config_init.go
+++ b/pkg/ck8s/config_init.go
@@ -78,12 +78,13 @@ func GenerateInitControlPlaneConfig(cfg InitControlPlaneConfig) (apiv1.Bootstrap
 		out.DatastoreServers = cfg.DatastoreServers
 	}
 
-	// CAPI manages the lifecycle of the Kubernetes node.
-	// Hence, k8s-snap should only clean up microcluster and files on upgrades.
-	out.ShouldRemoveK8sNode = ptr.To(false)
-
 	// annotations
 	out.ClusterConfig.Annotations = cfg.InitConfig.Annotations
+
+	// Since CAPI handles the lifecycle management of Kubernetes nodes, k8s-snap should only focus on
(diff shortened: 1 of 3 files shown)


pkg/ck8s/config_init.go
```diff
@@ -81,6 +81,11 @@ func GenerateInitControlPlaneConfig(cfg InitControlPlaneConfig) (apiv1.Bootstrap
 	// annotations
 	out.ClusterConfig.Annotations = cfg.InitConfig.Annotations
 
+	// Since CAPI handles the lifecycle management of Kubernetes nodes, k8s-snap should only focus on
+	// cleaning up microcluster and files during upgrades. This is essential for CAPI to function correctly.
+	// Therefore, instead of relying on the user to set this annotation in the initConfig, we hardcode it here.
+	out.ClusterConfig.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove] = "true"
```
neoaggelos, line 87: i'd recommend letting the user have the final say here, with something like:

```go
if _, ok := out.ClusterConfig.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove]; !ok {
   out.ClusterConfig.Annotations[apiv1.AnnotationSkipCleanupKubernetesNodeOnRemove] = "true"
}
```

neoaggelos (approved):

PR #37, opened 2024-08-05
Added initial in place update machine controller implementation

berkayoz:
Added initial in place update machine controller implementation

TODO:
* Authentication for the snap refresh endpoint

### DIFF
--- a/bootstrap/controllers/upgrade_controller.go
+++ b/bootstrap/controllers/upgrade_controller.go
@@ -1,0 +1,178 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/patch"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	bootstrapv1 "github.com/canonical/cluster-api-k8s/bootstrap/api/v1beta2"
+	"github.com/canonical/cluster-api-k8s/pkg/ck8s"
+)
+
+// InPlaceUpgradeReconciler reconciles machines and performs in-place upgrades based on annotations
+type InPlaceUpgradeReconciler struct {
+	client.Client
+	Log      logr.Logger
+	Scheme   *runtime.Scheme
+	recorder record.EventRecorder
+
+	K8sdDialTimeout time.Duration
+
+	managementCluster ck8s.ManagementCluster
+}
+
+func (r *InPlaceUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&clusterv1.Machine{}).
+		Build(r)
+	if err != nil {
+		return fmt.Errorf("failed setting up with a controller manager: %w", err)
+	}
+
+	r.Scheme = mgr.GetScheme()
+	r.recorder = mgr.GetEventRecorderFor("ck8s-in-place-upgrade-controller")
+
+	if r.managementCluster == nil {
+		r.managementCluster = &ck8s.Management{
+			Client:          r.Client,
+			K8sdDialTimeout: r.K8sdDialTimeout,
+		}
+	}
+	return err
+}
+
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;machines/status,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=bootstrap.cluster.x-k8s.io,resources=ck8sconfigs,verbs=get;list;watch
+// +kubebuilder:rbac:groups=bootstrap.cluster.x-k8s.io,resources=ck8sconfigs/status,verbs=get;list;watch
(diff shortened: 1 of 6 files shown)


pkg/ck8s/upgrade.go
```diff
@@ -0,0 +1,19 @@
+package ck8s
+
+const (
+	InPlaceUpgradeToAnnotation      = "k8sd.io/in-place-upgrade-to"
```
neoaggelos, line 4: move these annotations next to the existing annotations

templates/docker/Dockerfile
```diff
@@ -84,6 +84,11 @@ RUN /src/k8s-snap/build-scripts/build-component.sh containerd
 FROM builder AS build-helm
 RUN /src/k8s-snap/build-scripts/build-component.sh helm
 
+# # kubernetes build with older version for in place upgrade testing
+# FROM builder AS build-kubernetes-old
+# RUN echo "$(cat /src/k8s-snap/build-scripts/components/kubernetes/version | cut -f-2 -d'.').0" > /src/k8s-snap/build-scripts/components/kubernetes/version
+# RUN /src/k8s-snap/build-scripts/build-component.sh kubernetes
+
```
neoaggelos, line 91: we have (or will soon have) a `KUBERNETES_VERSION` arg to choose which Kubernetes version to build. Let's add a `KUBERNETES_VERSION_UPGRADE_TO`. If set, this builds an extra binary and adds it to our well-known location

neoaggelos (commented):
First pass, great work!

pkg/ck8s/workload_cluster.go
```diff
@@ -300,6 +355,58 @@ func (w *Workload) doK8sdRequest(ctx context.Context, method, endpoint string, r
...
+	if err := json.NewDecoder(res.Body).Decode(&responseBody); err != nil {
+		return fmt.Errorf("failed to parse HTTP response: %w", err)
+	}
+	if res.StatusCode != http.StatusOK {
+		return fmt.Errorf("HTTP request failed with status code: %d (%s)", res.StatusCode, responseBody.Error)
+	}
+	if responseBody.Error != "" {
+		return fmt.Errorf("k8sd request failed: %s", responseBody.Error)
+	}
+	if responseBody.Metadata == nil {
```
HomayoonAlimohammadi, line 399: let's also add this, otherwise we might encounter an error on `json.Unmarshal` if a nil response is passed (we had the same problem in `doK8sdRequest` for controlplane nodes).
```suggestion
	if responseBody.Metadata == nil || response == nil {
```

bootstrap/controllers/upgrade_controller.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		// Error reading the object - requeue the request.
+		return ctrl.Result{}, err
+	}
+
+	mAnnotations := m.GetAnnotations()
+
+	// Check if the machine has an in-place upgrade annotation
+	if upgradeOption, ok := mAnnotations[bootstrapv1.InPlaceUpgradeToAnnotation]; ok {
```
HomayoonAlimohammadi, line 81: nit: I feel like this if block up until the end of this function is very complex and excessively branched, maybe we can break this down so it's easier to read.
this is a completely a personal preference, feel free to ignore this.

pkg/ck8s/workload_cluster.go
```diff
@@ -204,6 +205,60 @@ func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context, options k8sd
...
+		return "", fmt.Errorf("failed to refresh machine %s: %w", machine.Name, err)
+	}
+
+	return response.ChangeID, nil
+}
+
+func (w *Workload) GetRefreshStatusForMachine(ctx context.Context, machine *clusterv1.Machine, nodeToken *string, changeID *string) (*apiv1.SnapRefreshStatusResponse, error) {
+	request := apiv1.SnapRefreshStatusRequest{}
+	response := &apiv1.SnapRefreshStatusResponse{}
+	request.ChangeID = *changeID
```
HomayoonAlimohammadi, line 248: nit: IMO it's better to keep the `changeID` every where, since we expect a `change ID` from snap (also consider changing the annotations). but I get that we're doing a `refresh` which has a `changeID` and is confusing by itself.

HomayoonAlimohammadi (commented):
Impressive work @berkayoz! Thanks a lot. left some small comments.

bootstrap/controllers/upgrade_controller.go
```diff
@@ -0,0 +1,245 @@
...
+			switch upgradeStatus {
+			case bootstrapv1.InPlaceUpgradeInProgressStatus:
+
+				status, err := workloadCluster.GetRefreshStatusForMachine(ctx, m, nodeToken, &refreshID)
+				if err != nil {
+					log.Info("Failed to get refresh status for machine", "error", err)
+					return ctrl.Result{}, errors.Wrapf(err, "failed to get refresh status for machine")
+				}
+
+				if !status.Completed {
```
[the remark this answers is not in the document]
berkayoz: This is coming from the snapd api, check the relevant comment on https://github.com/canonical/k8s-snap-api/pull/7#discussion_r1744860113

bootstrap/controllers/upgrade_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	K8sdDialTimeout time.Duration
+
+	managementCluster ck8s.ManagementCluster
+}
+
+func (r *InPlaceUpgradeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	_, err := ctrl.NewControllerManagedBy(mgr).
+		For(&clusterv1.Machine{}).
+		Build(r)
+	if err != nil {
```
bschimke95, line 42: nit: can be inlined

bootstrap/controllers/upgrade_controller.go
bschimke95: I'd just return if `!ok`

pkg/ck8s/workload_cluster.go
```diff
@@ -300,6 +355,58 @@ func (w *Workload) doK8sdRequest(ctx context.Context, method, endpoint string, r
 	return nil
 }
 
+func (w *Workload) doK8sdRequestForMachine(ctx context.Context, machine *clusterv1.Machine, method, endpoint string, header map[string][]string, request any, response any) error {
```
bschimke95, line 358: Hm, maybe refactor the already existing `doK8sdRequest` instead of duplicating the logic here?

bschimke95 (commented):
Great work, only minor comments then we should be good to merge.

PR #41, opened 2024-08-15
use hardcoded Go version to get go from go.dev/dl in Dockerfile

eaudetcobello:
old pattern could result in the following match which would cause a 404 during a docker build:

❯ curl -s https://go.dev/dl/ | grep -o "go1.22.*.linux-amd64.tar.gz"
go1.22.6.linux-amd64.tar.gz">go1.22.6.linux-amd64.tar.gz

### DIFF
--- a/templates/docker/Dockerfile
+++ b/templates/docker/Dockerfile
@@ -44,7 +44,7 @@ RUN apt-get update \
 ## NOTE(neoaggelos): Go version used to build components
 ## !!!IMPORTANT!!! Keep up to date with "snapcraft.yaml:parts.build-deps.build-snaps.go"
 ADD install-go.sh /
-RUN /install-go.sh 1.22 && ln -s /usr/local/go/bin/go /usr/local/bin/go
+RUN /install-go.sh && ln -s /usr/local/go/bin/go /usr/local/bin/go
 
 ## Prepare build environment
 ENV SNAPCRAFT_PART_INSTALL=/out
--- a/templates/docker/Dockerfile
+++ b/templates/docker/Dockerfile
@@ -44,7 +44,7 @@ RUN apt-get update \
 ## NOTE(neoaggelos): Go version used to build components
 ## !!!IMPORTANT!!! Keep up to date with "snapcraft.yaml:parts.build-deps.build-snaps.go"
 ADD install-go.sh /
-RUN /install-go.sh && ln -s /usr/local/go/bin/go /usr/local/bin/go
+RUN /install-go.sh 1.22.6 && ln -s /usr/local/go/bin/go /usr/local/bin/go
 
 ## Prepare build environment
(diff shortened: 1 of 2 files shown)


templates/docker/install-go.sh
```diff
@@ -9,7 +9,7 @@
 
 VERSION="$1"
 
-fname="$(curl -s https://go.dev/dl/ | grep -o "go$VERSION.*.linux-amd64.tar.gz" | head -1)"
+fname="$(curl -s https://go.dev/dl/ | grep -oP "go$VERSION\.\d+\.linux-amd64\.tar\.gz" | head -1)"
```
addyess, line 12: hows this?  Remove that perl grep stuff

```suggestion
fname="$(curl -s https://go.dev/dl/ | grep -oE "go$VERSION.[0-9]+.linux-amd64.tar.gz" | head -1)"
```

addyess (changes requested):
How's this @eaudetcobello ?

templates/docker/install-go.sh
addyess:
```bash
⚡ curl -s https://go.dev/dl/ | grep -o "go1.22.*.linux-amd64.tar.gz" | head -1
go1.22.6.linux-amd64.tar.gz">go1.22.6.linux-amd64.tar.gz
py
addyess@addyess-dell /home/addyess/git/charmed-kubernetes/jenkins                                                                                                      main
⚡ curl -s https://go.dev/dl/ | grep -oE "go1.22.[0-9]+.linux-amd64.tar.gz" | head -1
go1.22.6.linux-amd64.tar.gz
py
```

neoaggelos: the point is to keep it simple. instead of this, perhaps just hardcodd the full go version eg 1.22.6 and avoid smart grep  stuff?

templates/docker/Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN apt-get update \
 ## NOTE(neoaggelos): Go version used to build components
 ## !!!IMPORTANT!!! Keep up to date with "snapcraft.yaml:parts.build-deps.build-snaps.go"
 ADD install-go.sh /
-RUN /install-go.sh 1.22 && ln -s /usr/local/go/bin/go /usr/local/bin/go
+RUN /install-go.sh && ln -s /usr/local/go/bin/go /usr/local/bin/go
```
neoaggelos, line 47: i'd expect this to read `./install.go 1.22.6`, so that changing the version is done over the Dockerfile, not inside the script itself

neoaggelos (approved):

templates/docker/Dockerfile
eaudetcobello: Done

addyess (approved):

PR #44, opened 2024-08-19
Allow setting the node name in CK8sConfigTemplate

neoaggelos:
### Summary

Some cloud providers (e.g. AWS) require that the hostname be set to a specific value, typically `{{ ds.meta_data.local_hostname }}`.

Allow setting this field in the template, and seed this into the generated cloud-init.

### Note

`{{ ds.meta_data.local_hostname }}` will be passed to the cloud-init, then will be replaced with the desired value when Jinja renders it on the local node.


bschimke95 (approved):
LGTM, please fix the go test before merge

PR #22, opened 2024-07-08
AWS infrastructure provider template

bschimke95:

PR #45, opened 2024-08-19
Support running e2e tests on AWS

eaudetcobello:
cluster_upgrade_test.go → Here we replace the hardcoded reference to docker infrastructure and instead tell the test suite to resolve `. This is then read from the config. file of the test (I think), which lives at test/config/ck8s-{docker,aws}.yaml

create_test.go → same as above.

test/e2e/data/infrastructure-aws folder → this folder contains the templates that will be applied to the cluster during the test.

test/e2e/shared/v1beta1 → this defines the major and minor contract version of the core cluster api components that we use for the tests. (See ck8s-aws.yaml line 25)

test/e2e/shared/v1beta1_aws → this defines the major and minor contract version of the CAPA resources we deploy. (See ck8s-aws.yaml Line 38)

kcp_remediation_test.go, md_remediation_test.go, node_scale_test.go → Changed the hardcoded docker provider to clusterctl.DefaultInfrastructureProvider. Also, skip remediation tests on cloud provider.

Makefile -> added E2E_INFRA variable that can be specified like `E2E_INFRA=aws make test-e2e`


bschimke95 (commented):
great work @eaudetcobello, added couple of comments
we should probably get #22 in before this.

Makefile
```diff
@@ -92,14 +92,14 @@ KUSTOMIZE := $(TOOLS_BIN_DIR)/$(KUSTOMIZE_BIN)-$(KUSTOMIZE_VER)
...
+GINKGO_FOCUS ?= Workload cluster creation
 GINKGO_SKIP ?=
 GINKGO_NODES ?= 1 # GINKGO_NODES is the number of parallel nodes to run
                   # when running the e2e tests, 1 means no parallelism
 GINKGO_TIMEOUT ?= 2h
 GINKGO_POLL_PROGRESS_AFTER ?= 60m
 GINKGO_POLL_PROGRESS_INTERVAL ?= 5m
-E2E_CONF_FILE ?= $(TEST_DIR)/e2e/config/ck8s-docker.yaml
+E2E_CONF_FILE ?= $(TEST_DIR)/e2e/config/ck8s-aws.yaml
 SKIP_RESOURCE_CLEANUP ?= false
```
neoaggelos, line 103: this should be config, e.g. `E2E_INFRA ?= docker`, which users can set to `aws`

test/e2e/e2e_suite_test.go
```diff
@@ -196,6 +197,7 @@ func initScheme() *runtime.Scheme {
 	Expect(controlplanev1.AddToScheme(sc)).To(Succeed())
 	Expect(bootstrapv1.AddToScheme(sc)).To(Succeed())
 	Expect(dockerinfrav1.AddToScheme(sc)).To(Succeed())
+	Expect(awsinfrav1.AddToScheme(sc)).To(Succeed())
```
neoaggelos, line 200: we add the API, but do we use it anywhere? it brings in quite a few dependencies in our go.mod, let's be mindful of those

test/e2e/data/infrastructure-aws/cluster-template.yaml
neoaggelos, line 1: This needs to wait until #22

test/e2e/create_test.go
```diff
@@ -48,7 +48,7 @@ var _ = Describe("Workload cluster creation", func() {
 		Expect(e2eConfig.Variables).To(HaveKey(KubernetesVersion))
 
 		clusterName = fmt.Sprintf("capick8s-create-%s", util.RandomString(6))
-		infrastructureProvider = "docker"
```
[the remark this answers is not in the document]
eaudetcobello: I made it so the default infra is docker, and then the tests use clusterctl.DefaultInfrastructureProvider (not necessarily docker, it detects the installed provider).

hack/ci-e2e-tests.sh
```diff
@@ -0,0 +1,171 @@
...
+}
+
+function main {
+  if [[ $INFRA_PROVIDER != "aws" ]]; then
+    error_exit "Unsupported infrastructure provider: $INFRA_PROVIDER"
+    exit 1
+  fi
+
+  check_required_env_vars
+  install_lxd
```
HomayoonAlimohammadi, line 161: maybe we can check for lxd before installing it.

hack/ci-e2e-tests.sh
```diff
@@ -0,0 +1,171 @@
...
+  exec_in_container "clusterctl init -i $INFRA_PROVIDER -b ck8s:$CK8S_PROVIDER_VERSION -c ck8s:$CK8S_PROVIDER_VERSION --config /root/.cluster-api/clusterctl.yaml"
+}
+
+function run_e2e_tests {
+  make USE_EXISTING_CLUSTER=true GINKGO_FOCUS="Workload cluster creation" test-e2e
+}
+
+function cleanup {
+  if [[ $SKIP_CLEANUP == "true" ]]; then
+    return
```
HomayoonAlimohammadi, line 141: maybe we can also print something here.

hack/ci-e2e-tests.sh
```diff
@@ -0,0 +1,171 @@
...
+function error_exit {
+  printf "ERROR: %s\n" "$1" >&2
+  return 1
+}
+
+# Check that all required environment variables are set
+function check_required_env_vars {
+  local required_env_vars=()
+
+  if [[ $INFRA_PROVIDER == "aws" ]]; then
```
HomayoonAlimohammadi, line 39: so it's definitely over engineering but I was thinking about having an `InfraManager` interface which kicks in whenever it's needed, cause these if statements for different infras might start to get bloated. python would be a better option for these kinds of things, but imagine something like this:
```go
type InfraManager interface {
    CheckRequiredVars() error
    ConfigureContainerEnv() error
    InstallTools() error
    CleanUp() error
}

// implements InfraManager
type AWSInfraManager struct {
    // env vars and other config
    ...
}
```

HomayoonAlimohammadi: and then also we might have a map of available infras to their manager:
```go
var infraToManager = map[string]InfraManager{
    "aws": NewAWSInfraManager(config),
}
```

HomayoonAlimohammadi (commented):
Wonderful job @eaudetcobello! reviewed the `ci-e2e-tests.sh` and left some comments.

Makefile
```diff
@@ -92,14 +92,14 @@ KUSTOMIZE := $(TOOLS_BIN_DIR)/$(KUSTOMIZE_BIN)-$(KUSTOMIZE_VER)
 # Ginkgo
 TEST_DIR := $(shell pwd)/test
 ARTIFACTS ?= $(shell pwd)/_artifacts
-GINKGO_FOCUS ?=
+GINKGO_FOCUS ?= Workload cluster creation
```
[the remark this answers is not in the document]
eaudetcobello: Yes, this is temporary. I am leaving it like this for now because this test passes on all infras.

test/e2e/config/ck8s-aws.yaml
```diff
@@ -0,0 +1,118 @@
...
+          - sourcePath: "../data/shared/v1beta1/metadata-aws.yaml"
+        replacements:
+          - old: "imagePullPolicy: Always"
+            new: "imagePullPolicy: IfNotPresent"
+
+      # Add v1.7.99 to support tilt (not presented in ../data/shared/v1beta1/metadata.yaml)
+      # when bootstrapping with tilt, it will use
+      # the defaultProviderVersion in https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/tools/internal/tilt-prepare/main.go as
+      # default version for docker infrastructure provider
+      # name here should match defaultProviderVersion
```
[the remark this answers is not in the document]
eaudetcobello: It belongs, but I will clean up the message.

hack/ci-e2e-tests.sh
eaudetcobello: I'm thinking of removing the cleanup function from the script, and having the user run that as a separate step. It reduces the size and complexity of the script a bit.

hack/ci-e2e-tests.sh
```diff
@@ -0,0 +1,171 @@
+#!/bin/bash
+
+set -xe
+
+# This script is used to run e2e tests for the CK8s CAPI.
+# It sets up an LXD container, installs the CK8s management cluster, and runs e2e tests.
+# The goal is to test the CK8s provider with different infrastructure providers (e.g., AWS, Azure, GCP). Only AWS is supported for now.
```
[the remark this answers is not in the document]
eaudetcobello: I think I'm doing that in the main function.

hack/ci-e2e-tests.sh
eaudetcobello: Right. That's not a bad idea. We could rewrite this in a more structured language as a follow-up to this PR.

.github/workflows/e2e-deleteme.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Check out repo
+        uses: actions/checkout@v4
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        with:
+          detached: true
+      - name: Install requirements
+        run: |
+          sudo apt install make
+          sudo apt install wget
```
HomayoonAlimohammadi, line 39: maybe we can install everything in a single line

.github/workflows/e2e-deleteme.yaml
HomayoonAlimohammadi, line 1: I'm a bit confused about the name of this file, I guess the contents of this file was supposed to be carried over to the `e2e.yaml` file or something

.github/workflows/e2e.yaml
```diff
@@ -2,6 +2,8 @@ name: E2E Tests
 
 on:
   pull_request:
+    branches:
+      - does-not-exist
```
HomayoonAlimohammadi, line 6: and I guess after moving the content of `deleteme` file to this file, we should remove this line

Makefile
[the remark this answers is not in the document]
HomayoonAlimohammadi: by temporary do we mean that it's gonna get removed before this PR is merged? or are we planning to change this in the up coming PRs? if it's the latter, maybe we can add a note or todo.
also I think this will change the scope for both the docker and aws tests, right? I think we only need this for the aws test.

hack/aws-nuke-config.yaml
```diff
@@ -0,0 +1,96 @@
+regions:
+  - us-east-2
```
HomayoonAlimohammadi, line 2: maybe we can extract this from this file and also the workflow file. I think this might cause subtle problems if we duplicate it.

.github/workflows/e2e-deleteme.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Increase inotify watches
+        run: |
+          # Prevents https://cluster-api.sigs.k8s.io/user/troubleshooting#cluster-api-with-docker----too-many-open-files
+          sudo sysctl fs.inotify.max_user_watches=1048576
+          sudo sysctl fs.inotify.max_user_instances=8192
+      - name: Run e2e tests
+        run: |
+          sudo -E ./hack/ci-e2e-tests.sh true aws v0.1.2
```
HomayoonAlimohammadi, line 47: I can't understand this `true` here, would elaborate a bit? I see `${0}`, `${2}` and `${3}` in the script, this should be `${1}`, right?

hack/ci-e2e-tests.sh
HomayoonAlimohammadi, line 1: really nice work putting up this script together. it should have been really challenging. but I strongly feel like we need to change this to a (maybe) python script cause it's too complicated for a bash script. not for this task but in the near future.

HomayoonAlimohammadi (commented):
Really nice work! Thanks a lot @eaudetcobello! did an initial pass and left some comments.

.github/workflows/e2e.yaml
```diff
@@ -24,20 +24,51 @@ jobs:
...
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-images
+          path: |
+            provider-images.tar
+
+  build-k8s-snap-e2e-images:
+    name: Build K8s Snap E2E Images
+    if: false
```
bschimke95, line 50: reminder to remove before merge.

.github/workflows/e2e.yaml
```diff
@@ -24,20 +24,51 @@ jobs:
...
+          sudo apt install -y make docker-buildx
           sudo snap install go --classic --channel=1.22/stable
-          sudo apt install make
-          sudo apt install docker-buildx
           sudo snap install kubectl --classic --channel=1.30/stable
       - name: Build provider images
-        run: sudo make docker-build-e2e
-      - name: Build k8s-snap images
-        working-directory: hack/
+        #run: sudo make docker-build-e2e
```
bschimke95, line 31: reminder to enable

templates/aws/cluster-generated.yaml
bschimke95, line 1: to be removed.

test/e2e/data/infrastructure-aws/cluster-template.yaml
```diff
@@ -0,0 +1,328 @@
...
+spec:
+  region: ${AWS_REGION}
+  sshKeyName: ${AWS_SSH_KEY_NAME}
+  bastion:
+    enabled: ${AWS_CREATE_BASTION}
+  controlPlaneLoadBalancer:
+    healthCheckProtocol: TCP
+  network:
+    vpc:
+      id: vpc-09dc405df98c78806
```
bschimke95, line 37: does this mean we always have a fixed VPC deployed and assume that it exists?

test/e2e/kcp_remediation_test.go
```diff
@@ -23,16 +23,25 @@ import (
...
 	capi_e2e "sigs.k8s.io/cluster-api/test/e2e"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
 )
 
 var _ = Describe("When testing KCP remediation", func() {
+	if clusterctl.DefaultInfrastructureProvider == "aws" {
+		// Skip the test for cloud provider as it is not supported
+		// See kubernetes.slack.com/archives/C8TSNPY4T/p1680525266510109
+		// And github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4198
+		return
```
bschimke95, line 34: Ginko also allows to explicitly skip tests (`ginko.Skip()` or something like this)

bschimke95 (commented):
The card is in review but I guess there is some cleanup work to be done. 
Added some comments that you are probably already aware of.

test/e2e/data/infrastructure-aws/cluster-template.yaml
eaudetcobello: I needed this to try to set a security group that allows all incoming traffic. Cannot create a security group without specifying a VPC, so I created one.

Once we find the root cause of the networking issue we can remove this.

.vscode/settings.json
bschimke95, line 1: I guess you didn't want to share your VSCode colors with us.

.github/workflows/e2e.yaml
```diff
@@ -109,4 +109,4 @@ jobs:
           sudo sysctl fs.inotify.max_user_instances=8192
       - name: Run e2e tests
         run: |
-          sudo GINKGO_FOCUS="${{ matrix.ginkgo_focus }}" SKIP_RESOURCE_CLEANUP=true make test-e2e
+          sudo E2E_INFRA=docker GINKGO_FOCUS="${{ matrix.ginkgo_focus }}" SKIP_RESOURCE_CLEANUP=true make test-e2e
```
bschimke95, line 112: looks like `docker` is the default, so this is not required.

test/e2e/config/ck8s-aws.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+variables:
+  KUBERNETES_VERSION_MANAGEMENT: "v1.28.0"
+  KUBERNETES_VERSION: "v1.30.0"
+  KUBERNETES_VERSION_UPGRADE_TO: "v1.30.1"
+  IP_FAMILY: "IPv4"
+  KIND_IMAGE_VERSION: "v1.28.0"
+  AWS_CONTROL_PLANE_MACHINE_TYPE: t3.large
+  AWS_NODE_MACHINE_TYPE: t3.large
+  AWS_SSH_KEY_NAME: "etienne"
```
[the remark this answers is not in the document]
bschimke95: Looks like this is not addressed

test/e2e/data/infrastructure-aws/cluster-template.yaml
```diff
@@ -0,0 +1,315 @@
...
+  namespace: default
+spec:
+  machineTemplate:
+    infrastructureTemplate:
+      apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
+      kind: AWSMachineTemplate
+      name: ${CLUSTER_NAME}-control-plane
+  spec:
+    nodeName: "{{ ds.meta_data.local_hostname }}"
+    channel: "1.31-classic/edge"
```
bschimke95, line 55: This should be "${KUBERNETES_VERSION}-classic/edge"

test/e2e/data/infrastructure-aws/cluster-template.yaml
```diff
@@ -0,0 +1,315 @@
...
+---
+apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
+kind: CK8sConfigTemplate
+metadata:
+  name: ${CLUSTER_NAME}-md-0
+spec:
+  template:
+    spec:
+      nodeName: "{{ ds.meta_data.local_hostname }}"
+      channel: "1.31-classic/edge"
```
bschimke95, line 125: same here

bschimke95 (commented):
Final comments, then we should be good to merge.

bschimke95 (approved):
LGTM, great stuff @eaudetcobello
Thanks for pushing through it.

PR #46, opened 2024-08-27
Use Node name instead of Machine name on removal

HomayoonAlimohammadi:
# Summary
This PR fixes the issue with CAPI rolling upgrades. The problem was that we were passing the wrong `node-name` to k8sd `remove-node` endpoint. 

# Description
CAPI rolling upgrades were failing, we narrowed down the problem, figured that the node removal process is not happening as expected (i.e. you can reproduce the problem by scaling down the control plane replicas of the workload cluster). Finally it was evident that the `/capi/remove-node` endpoint of k8sd was returning an error indicating that it can not find the node it was called to delete. That is because we were passing the `machine.Name` to k8sd `/capi/remove-node` endpoint, while as can be seen in the output of `kubectl get nodes` (of the workload cluster), we need to pass the node name.

### DIFF
--- a/pkg/ck8s/workload_cluster.go
+++ b/pkg/ck8s/workload_cluster.go
@@ -230,11 +230,18 @@ func (w *Workload) requestJoinToken(ctx context.Context, name string, worker boo
 }
 
 func (w *Workload) RemoveMachineFromCluster(ctx context.Context, machine *clusterv1.Machine) error {
-	request := &apiv1.RemoveNodeRequest{Name: machine.Name, Force: true}
+	if machine == nil {
+		return fmt.Errorf("machine is nil")
+	} else if machine.Status.NodeRef == nil {
+		return fmt.Errorf("machine %s has no node reference", machine.Name)
+	}
+
+	nodeName := machine.Status.NodeRef.Name
+	request := &apiv1.RemoveNodeRequest{Name: nodeName, Force: true}
 
 	// If we see that ignoring control-planes is causing issues, let's consider removing it.
 	// It *should* not be necessary as a machine should be able to remove itself from the cluster.
-	err := w.doK8sdRequest(ctx, http.MethodPost, "1.0/x/capi/remove-node", request, nil, k8sdProxyOptions{IgnoreNodes: map[string]struct{}{machine.Name: {}}})
+	err := w.doK8sdRequest(ctx, http.MethodPost, "1.0/x/capi/remove-node", request, nil, k8sdProxyOptions{IgnoreNodes: map[string]struct{}{nodeName: {}}})
 	if err != nil {
 		return fmt.Errorf("failed to remove %s from cluster: %w", machine.Name, err)
 	}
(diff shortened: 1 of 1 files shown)


addyess:
@HomayoonAlimohammadi i came to review this -- but if you could please add some description and references to why this change is necessary.  Sorry, i don't have enough context to see how this is an obvious change.  thanks

pkg/ck8s/workload_cluster.go
```diff
@@ -230,11 +230,18 @@ func (w *Workload) requestJoinToken(ctx context.Context, name string, worker boo
...
 
 func (w *Workload) RemoveMachineFromCluster(ctx context.Context, machine *clusterv1.Machine) error {
-	request := &apiv1.RemoveNodeRequest{Name: machine.Name, Force: true}
+	if machine == nil {
+		return fmt.Errorf("machine is nil")
+	} else if machine.Status.NodeRef == nil {
+		return fmt.Errorf("machine %s has no node reference", machine.Name)
+	}
+
+	nodeName := machine.Status.NodeRef.Name
```
eaudetcobello, line 239: Nice work finding the root cause. I only have one question, otherwise this looks good. Is Status.NodeRef set automatically by CAPI's Machine controller?

HomayoonAlimohammadi:
> @HomayoonAlimohammadi i came to review this -- but if you could please add some description and references to why this change is necessary. Sorry, i don't have enough context to see how this is an obvious change. thanks

I'm really sorry @addyess, you're right, the PR was rushed out of excitement! :D 
Tried to update the description accordingly, lemme know if I can elaborate further.

pkg/ck8s/workload_cluster.go
```diff
@@ -230,11 +230,18 @@ func (w *Workload) requestJoinToken(ctx context.Context, name string, worker boo
 }
 
 func (w *Workload) RemoveMachineFromCluster(ctx context.Context, machine *clusterv1.Machine) error {
-	request := &apiv1.RemoveNodeRequest{Name: machine.Name, Force: true}
+	if machine == nil {
+		return fmt.Errorf("machine is nil")
+	} else if machine.Status.NodeRef == nil {
```
bschimke95, line 235: nit: I'd split this into separate statements.
```
if condition1 {
  return ...
}
if condition2 {
  return ...
}
```
IMHO is slightly more readable.

bschimke95 (approved):
LGTM

pkg/ck8s/workload_cluster.go
HomayoonAlimohammadi: I think it does:
- https://cluster-api.sigs.k8s.io/developer/architecture/controllers/machine-pool#:~:text=Setting%20NodeRefs%20on%20MachinePool%20instances%20to%20be%20able%20to%20associate%20them%20with%20Kubernetes%20nodes
- https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/machine/machine_controller_noderef.go#L90-L100

bschimke95: yes, this is part of the CAPI contract.

PR #48, opened 2024-08-30
Fix annotation, Skip nil response

HomayoonAlimohammadi:
We use `SkipCleanup` annotation to make sure k8sd does not remove the actual k8s node when calling `/x/capi/remove-node` endpoint so that CAPI core provider can take care of that. This PR makes sure the annotation is available in the cluster config everytime (currently we might end up not without the annotation).
Also there was a minor issue in the `doK8sdRequest` in which we tried to unmarshal metadata in a nil response.


bschimke95 (approved):
LGTM

PR #54, opened 2024-09-09
Add Certs Refresh Proposal

mateoflorido:
## Overview
This proposal outlines a plan for refreshing certificates on cluster machines managed by Canonical Kubernetes CAPI providers.

### DIFF
--- a/docs/002-refresh-certs.md
+++ b/docs/002-refresh-certs.md
@@ -1,0 +1,347 @@
+<!--
+To start a new proposal, create a copy of this template on this directory and
+fill out the sections below.
+-->
+
+# Proposal information
+
+<!-- Index number -->
+- **Index**: 002
+
+<!-- Status -->
+- **Status**: **DRAFTING**
+<!-- **DRAFTING**/**ACCEPTED**/**REJECTED** -->
+
+<!-- Short description for the feature -->
+- **Name**: ClusterAPI Certificates Refresh
+
+<!-- Owner name and github handle -->
+- **Owner**: Mateo Florido [@mateoflorido](https://github.com/mateoflorido)
+<!-- [@name](https://github.com/name) -->
+
+# Proposal Details
+
+## Summary
+<!--
+In a short paragraph, explain what the proposal is about and what problem
+it is attempting to solve.
+-->
+
+The proposal aims to enhance Canonical Kubernetes Cluster API Providers by
+enabling administrators to refresh or renew certificates on cluster nodes
+without the need for a rolling upgrade. This feature is particularly beneficial
+in resource-constrained environments, such as private or edge clouds, where
+performing a full node replacement may not be feasible.
+
+## Rationale
+<!--
+This section COULD be as short or as long as needed. In the appropriate amount
+of detail, you SHOULD explain how this proposal improves k8s providers, what is the
+problem it is trying to solve and how this makes the user experience better.
+
+You can do this by describing user scenarios, and how this feature helps them.
+You can also provide examples of how this feature may be used.
+-->
+
+Currently, Cluster API lacks a mechanism for refreshing certificates on cluster
+nodes without triggering a full rolling update. For example, while the Kubeadm
+provider offers the ability to renew certificates, it requires a rolling update
+of the cluster nodes or manual intervention before the certificates expire.
+
+This proposal aims to address this gap by enabling certificate renewal on
+cluster nodes without requiring a rolling update. By providing administrators
+with the ability to refresh certificates independently of node upgrades, this
+feature improves cluster operation, especially in environments with limited
+resources, such as private or edge clouds.
+
+It will enhance the user experience by minimizing downtime, reducing the need
+for additional resources, and simplifying certificate management. This is
+particularly valuable for users who need to maintain continuous availability
+or operate in environments where rolling updates are not practical due to
+resource constraints.
+
+
+## User facing changes
+<!--
+This section MUST describe any user-facing changes that this feature brings, if
+any. If an API change is required, the affected endpoints MUST be mentioned. If
+the output of any k8s command changes, the difference MUST be mentioned, with a
+clear example of "before" and "after".
+-->
+
+Administrators will be able to renew certificates on cluster nodes without
+triggering a full rolling update. This can be achieved by annotating the Machine
+object, which will initiate the certificate renewal process:
+
+```
+kubectl annotate machine <machine-name> k8sd.io/refresh-certificates={expires-in}
+```
+
+`expires-in` specifies how long the certificate will remain valid. It can be
+expressed in years, months, days, or any other time unit supported by the
+`time.ParseDuration`.
+
+For tracking the validity of certificates, the Machine object will include a
+`machine.cluster.x-k8s.io/certificates-expiry` annotation that indicates the
+expiry date of the certificates. This annotation will be added when the cluster
+is deployed and updated when certificates are renewed. The value of this
+annotation will be a RFC3339 timestamp.
+
+## Alternative solutions
+<!--
+This section SHOULD list any possible alternative solutions that have been or
+should be considered. If required, add more details about why these alternative
+solutions were discarded.
+-->
+
+**Kubeadm Control Plane provider (KCP)** automates certificate rotations for
+control plane machines by triggering a machine rollout when certificates are
+close to expiration.
+
+### How to configure:
+- In the KCP configuration, set the `rolloutBefore.certificatesExpiryDays`
+field. This tells KCP when to trigger the rollout before certificates expire:
+
+```yaml
+spec:
+  rolloutBefore:
+    certificatesExpiryDays: 21  # Trigger rollout when certificates expire within 21 days
+```
+
+### How it works:
+- **Automatic Rollouts**: KCP monitors the certificate expiry dates of control
+plane machines using the `Machine.Status.CertificatesExpiryDate`. If
+certificates are about to expire (based on a configured threshold), KCP
+triggers a machine rollout to refresh them.
+- **Certificate Expiry Check**: The expiry date is sourced from the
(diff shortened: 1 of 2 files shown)


docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+<!--
+This section COULD be as short or as long as needed. In the appropriate amount
+of detail, you SHOULD explain how this proposal improves k8s providers, what is the
+problem it is trying to solve and how this makes the user experience better.
+
+You can do this by describing user scenarios, and how this feature helps them.
+You can also provide examples of how this feature may be used.
+-->
+
+Currently, Cluster API lacks a mechanism for refreshing certificates on cluster
```
louiseschmidtgen, line 46: Have we considered an upstream contribution?

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+}
+```
+
+## Bootstrap Provider Changes
+<!--
+This section MUST mention any changes to the bootstrap provider.
+-->
+
+A controller called `CertificatesController` will be added to the bootstrap
+provider. This controller will watch for the `k8sd.io/refresh-certificates`
```
louiseschmidtgen, line 224: When you say watch here is this a kubernetes watch or rather polling this endpoint `GET /x/capi/certificates-expiry` every day?
I'm saying this with the thought in the back of my mind that watchers are expensive in terms of the resource they use. I don't think this would be relevant for small clusters but if we think about the 25k node scenario then we need to think about it.

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+
+The controller will share the CSR names with the control plane node by updating
+the `k8sd.io/refresh-certificates-csr-names` annotation with the list of CSR names
+to approve. The control plane node will then call the `POST /x/capi/approve-certificates`
+endpoint to approve the Certificate Signing Requests list provided by the annotation
+value.
+
+The controller will share the status similar to the control plane nodes by updating
+the `k8sd.io/refresh-certificates-status` annotation with the status of the renewal.
+The value of this annotation will be the same as the control plane nodes.
```
louiseschmidtgen, line 267: Thinking about edge cases, what are we going to do when a node gets stuck in `in progress`. Do we timeout and try again after some time?

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+close to expiration.
+
+### How to configure:
+- In the KCP configuration, set the `rolloutBefore.certificatesExpiryDays`
+field. This tells KCP when to trigger the rollout before certificates expire:
+
+```yaml
+spec:
+  rolloutBefore:
+    certificatesExpiryDays: 21  # Trigger rollout when certificates expire within 21 days
```
louiseschmidtgen, line 108: Is 21 days our default value? What are the reasons for choosing 21 days?

louiseschmidtgen (commented):
You've put some great thought into this cert refresh proposal, thank you Mateo!
Posted a few questions that came to my mind while reviewing.

docs/002-refresh-certs.md
berkayoz: A `Controller` in CAPI always watches a type of Kubernetes object. For this proposal it will be a `Machine`, so when the annotations are added controller will run a reconcile function to handle the changes. The watch operation is not avoidable here as it's the core concept.

docs/002-refresh-certs.md
berkayoz: This is the upstream value, we are not choosing a default value here.

docs/002-refresh-certs.md
berkayoz: I think this is not very feasible under the upstream assumption of immutable infrastructure, we could work towards an upstream contribution after we deliver an initial solution. Even currently the certificate refreshes are provider based, the default capi controllers do not have any mechanism for this.

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+The controller would use the value of the `k8sd.io/refresh-certificates` annotation
+to determine the duration after which the certificates will expire. It will then
+call the `POST /x/capi/request-certificates` endpoint to create the Certificate
+Signing Request (CSR) for the worker node.
+
+The controller will share the CSR names with the control plane node by updating
+the `k8sd.io/refresh-certificates-csr-names` annotation with the list of CSR names
+to approve. The control plane node will then call the `POST /x/capi/approve-certificates`
+endpoint to approve the Certificate Signing Requests list provided by the annotation
+value.
```
berkayoz, line 263: I'm not sure if we need to annotate control plane machines, we have the k8sd proxy which we can use to talk to the k8sd  of a control plane node directly. I assume the provider can just use this proxy to request the CSR's to be approved.

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+the output of any k8s command changes, the difference MUST be mentioned, with a
+clear example of "before" and "after".
+-->
+
+Administrators will be able to renew certificates on cluster nodes without
+triggering a full rolling update. This can be achieved by annotating the Machine
+object, which will initiate the certificate renewal process:
+
+```
+kubectl annotate machine <machine-name> k8sd.io/refresh-certificates={expires-in}
```
berkayoz, line 77: I think we should version our annotations, we've done `v1beta2.k8sd.io/*` to match the API version.

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+renewal process.
+
+The controller will share the status of the certificate renewal process with the
+Machine object by updating the `k8sd.io/refresh-certificates-status` annotation
+with the status of the renewal process. The value of this annotation will be
+one of the following:
+
+- `in-progress`: The certificate renewal process is in progress.
+- `done`: The certificate renewal process is complete.
+- `failed`: The certificate renewal process has failed.
```
berkayoz, line 242: We should add `Events` for each step and provide the failure message through here.

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+feature. Most features are expected to come with at least a How-To and an
+Explanation page.
+
+In this section, it is useful to think about any existing pages that need to be
+updated (e.g. command outputs).
+-->
+
+This implementation will require adding the following documentation:
+- How-to guide for renewing certificates on cluster nodes
+- Reference page of the `k8sd.io/renew-certificates` annotation
```
berkayoz, line 308: Are we doing `k8sd.io/renew-certificates` or `k8sd.io/refresh-certificates`?

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+### `GET /x/capi/certificates-expiry`
+
+This endpoint will return the expiry date of the certificates on a specific
+cluster node. The response will include the expiry date of the certificates
+in RFC3339 format.
+
+```go
+type CertificatesExpiryResponse struct {
+  // ExpiryDate is the expiry date of the certificates on the node.
+  ExpiryDate string `json:"expiry-date"`
```
bschimke95, line 167: What if they have different expiry dates? e.g. you can bring your own certificates as part of the bootstrap config or is this out of scope?

bschimke95: Maybe use `ExpiryDate is the expiry date of the earliest expiring certificate on the node`.
That way, we can have a `/x/capi/certificates-expiry/<certificate-name>` in the future without API changes.

docs/002-refresh-certs.md
```diff
@@ -0,0 +1,347 @@
...
+<!--
+This section MUST mention any changes to the k8sd API, or any additional API
+endpoints (and messages) that are required for this proposal.
+
+Unless there is a particularly strong reason, it is preferable to add new v2/v3
+APIs endpoints instead of breaking the existing APIs, such that API clients are
+not affected.
+-->
+
+### `GET /x/capi/certificates-expiry`
```
bschimke95, line 158: Should we really have this in the `/x/capi` namespace? This sounds like something that could be handy in other use-cases as well.

docs/002-refresh-certs.md
mateoflorido: We can handle cases where the request times out or fails by requeuing the event and retrying after a short delay.

docs/002-refresh-certs.md
mateoflorido: Yes, that's an extract from the upstream CAPI book: https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp#configuring-machine-rollout

docs/002-refresh-certs.md
mateoflorido: Providing your own certificates should be out of scope. In the upstream implementation of Kubeadm, the API server certificate's expiration date is used for the annotation value. It's uncommon to have different expiration dates for certificates on the same node: https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp#certificate-expiry-time

docs/002-refresh-certs.md
mateoflorido: I think it could be useful for other functionalities (like the status command). If that's something we would want to implement, we should move it out the `x/capi` namespace.

bschimke95: Let's move it out of that namespace.

docs/002-refresh-certs.md
bschimke95: ack - thanks for the context.

bschimke95 (approved):
LGTM, great work!

berkayoz (approved):
Great work! Let's mark it as accepted before merging

bschimke95 (approved):
LGTM

PR #56, opened 2024-09-15
Adapt scale test to not remove HA

bschimke95:
This PR fixes a race condition that pops up if a workload cluster is scaled down. As part of the CK8s node removal, all services are stopped on the leaving node. This includes `kubelet` which causes the node to be reported as "not available" before the core CAPI controllers attempt to drain the node. 

By setting the annotation, we ensure that CAPI takes full control of the nodes removal process.
Note that we could also change the existing `skip-node-on-removal` annotation but there might be another use-case for just stopping services (+backward-compatibility).

Requires: 
https://github.com/canonical/k8s-snap/pull/682
https://github.com/canonical/k8s-snap-api/pull/8

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/canonical/cluster-api-k8s
 go 1.22.6
 
 require (
-	github.com/canonical/k8s-snap-api v1.0.5
+	github.com/canonical/k8s-snap-api v1.0.6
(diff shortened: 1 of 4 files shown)


berkayoz (approved):
LGTM

HomayoonAlimohammadi (approved):
Thanks a lot @bschimke95! LGTM

PR #58, opened 2024-09-26
Add Certificates Refresh Controller - Control Plane

mateoflorido:
> [!IMPORTANT]  
> This implementation depends on the work in https://github.com/canonical/k8s-snap/pull/699
## Overview
Create a new controller to handle refresh certificates for control plane nodes running on CAPI.
## Rationale
We need to expose the certificate refresh functionality for control plane nodes without requiring a full rollout deployment, which would consume more resources. Instead, we leverage the business logic baked in `k8s-snap` to refresh the certificates in place, which allows us to refresh the self-signed certificates without the additional overhead or down time in the cluster.
## Changes
- Implemented a new Certificates Controller.
- Added logic to annotate machines with the certificate expiration date, enabling the expiry date to bubble up from the node.
- Added functionality to refresh certificates for control plane nodes.
- Increased the default `socat` timeout (0.5 seconds) to 5 seconds due to latency when creating new certificates.
## Further Information
Proposal 003 - https://github.com/canonical/cluster-api-k8s/pull/54


bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if refreshCertificates {
+		if configOwner.IsControlPlaneMachine() {
+			return r.refreshControlPlaneCertificates(ctx, scope)
+		} else {
+			return ctrl.Result{}, fmt.Errorf("Worker nodes are not supported yet")
+		}
+	}
+
+	if !updateExpiryDate {
```
eaudetcobello, line 144: Why are we calling updateExpirtyDateAnnotation *if* updateExpiryDate is *false*?

bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,219 @@
...
+	extraSANs := append(controlPlaneConfig.ExtraSANs, controlPlaneEndpoint)
+
+	expirySecondsUnix, err := scope.Workload.RefreshCertificates(ctx, scope.Machine, *nodeToken, seconds, extraSANs)
+	if err != nil {
+		scope.Log.Error(err, "Failed to refresh certificates")
+		r.recorder.Eventf(scope.Machine, corev1.EventTypeWarning, bootstrapv1.CertificatesRefreshFailedEvent, "Failed to refresh certificates: %v", err)
+		return ctrl.Result{}, err
+	}
+
+	scope.Log.Info("Certificates refreshed", "expiry", expirySecondsUnix)
```
eaudetcobello, line 182:
```suggestion
scope.Log.Info("Certificates refreshed", 
               "machine", scope.Machine.Name, 
               "cluster", scope.Cluster.Name, 
               "expiry", expirySecondsUnix)
```

bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		// Machine is being deleted, return early.
+		return ctrl.Result{}, nil
+	}
+
+	mAnnotations := m.GetAnnotations()
+
+	var refreshCertificates, updateExpiryDate bool
+	_, refreshCertificates = mAnnotations[bootstrapv1.CertificatesRefreshAnnotation]
+	_, updateExpiryDate = mAnnotations[bootstrapv1.MachineCertificatesExpiryDateAnnotation]
+	if !refreshCertificates && updateExpiryDate {
```
eaudetcobello, line 93: I think you meant for both to be negative?
```suggestion
	if !refreshCertificates && !updateExpiryDate {
```

bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,219 @@
...
+	scope.Log.Info("Certificates refreshed", "expiry", expirySecondsUnix)
+
+	expiryTime := time.Unix(int64(expirySecondsUnix), 0)
+
+	delete(mAnnotations, bootstrapv1.CertificatesRefreshAnnotation)
+	mAnnotations[bootstrapv1.MachineCertificatesExpiryDateAnnotation] = expiryTime.Format(time.RFC3339)
+	scope.Machine.SetAnnotations(mAnnotations)
+	if err := scope.Patcher.Patch(ctx, scope.Machine); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to patch machine annotations: %w", err)
+	}
```
eaudetcobello, line 191: Just an idea... how about requeuing if this failed because of a concurrent update?

https://sourcegraph.com/search?q=context:global+apierrors.IsConflict&patternType=keyword&sm=0

bootstrap/controllers/certificates_controller.go
mateoflorido: I think the variable name is a bit misleading. I'll update it. TL;DR: This should check for the presence of the annotation on the machine. If the annotation is not there, we need to add it.

bootstrap/controllers/certificates_controller.go
mateoflorido: Please see this comment: https://github.com/canonical/cluster-api-k8s/pull/58#discussion_r1777384713

bootstrap/controllers/certificates_controller.go
mateoflorido: In this case, if the returned error is non-nil, the Result is ignored and the request will be requeued using exponential backoff. The only way to stop this from reconciling is to return a TerminalError, so in this controller, everything will be retried using the exponential backoff mechanism.

eaudetcobello (approved):

bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,229 @@
...
+
+	if refreshCertificates {
+		if configOwner.IsControlPlaneMachine() {
+			return r.refreshControlPlaneCertificates(ctx, scope)
+		} else {
+			return ctrl.Result{}, fmt.Errorf("worker nodes are not supported yet")
+		}
+	}
+
+	if !hasExpiryDateAnnotation {
```
berkayoz, line 148: I think this should take precadance to process first if the `refreshCertificates || !hasExpiryDateAnnotation` conditions are both met. The expiry should be updated first, then a certificate refresh request should be processed.

bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,229 @@
...
+		Cluster:  cluster,
+		Patcher:  patchHelper,
+		Workload: workload,
+	}
+
+	if refreshCertificates {
+		if configOwner.IsControlPlaneMachine() {
+			return r.refreshControlPlaneCertificates(ctx, scope)
+		} else {
+			return ctrl.Result{}, fmt.Errorf("worker nodes are not supported yet")
```
berkayoz, line 144: This would infinitely loop on a retry since an error is returned, we should log not supported and ignore the request by returning `ctrl.Result{}, nil`

pkg/ck8s/manifests/k8sd-proxy-template.yaml
```diff
@@ -45,8 +45,9 @@ spec:
               name: k8sd-proxy-config
               key: k8sd-port
         args:
-        - TCP4-LISTEN:2380,fork,reuseaddr
-        - TCP4:$(HOSTIP):$(K8SD_PORT)
+        - -t 5
```
berkayoz, line 48: Can we add a comment here on the error we've faced and what this resolves so we can refer it in the future.

berkayoz (commented):
Great work! Did a first pass and left some comments.

bootstrap/controllers/certificates_controller.go
```diff
@@ -0,0 +1,229 @@
...
+	scope := &CertificatesScope{
+		Log:      log,
+		Machine:  m,
+		Config:   config,
+		Cluster:  cluster,
+		Patcher:  patchHelper,
+		Workload: workload,
+	}
+
+	if refreshCertificates {
```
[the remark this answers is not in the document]
berkayoz: Ignore this, the operation is done on the same event the annotation won't be present for other events in the queue.

berkayoz (approved):

PR #59, opened 2024-09-26
rename max to maxDuration to fix predeclared lint

eaudetcobello:
Fixes: `controlplane/controllers/remediation.go:377:6: function max has same name as predeclared identifier (predeclared)`

### DIFF
--- a/controlplane/controllers/remediation.go
+++ b/controlplane/controllers/remediation.go
@@ -373,7 +373,7 @@ func (r *CK8sControlPlaneReconciler) checkRetryLimits(log logr.Logger, machineTo
 	return remediationInProgressData, true, nil
 }
 
-// maxDuration calculates the maximum duration.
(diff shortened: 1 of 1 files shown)


controlplane/controllers/remediation.go
```diff
@@ -373,8 +373,8 @@ func (r *CK8sControlPlaneReconciler) checkRetryLimits(log logr.Logger, machineTo
 	return remediationInProgressData, true, nil
 }
 
-// max calculates the maximum duration.
-func max(x, y time.Duration) time.Duration {
+// maxDuration calculates the maximum duration.
```
mateoflorido, line 376:
```suggestion
// maxDuration returns the longer of two time.Duration values.
```
I think the original godoc is also quite odd, because there's no calculation involved.

mateoflorido (approved):

PR #60, opened 2024-10-01
Add Cluster-wide Certificates Refresh

mateoflorido:
## Overview
Add a proposal for implementing a cluster-wide certificates refresh in the Canonical Kubernetes providers

### DIFF
--- a/docs/proposals/004-orchestration-refresh-certs.md
+++ b/docs/proposals/004-orchestration-refresh-certs.md
@@ -9,7 +9,7 @@ fill out the sections below.
 - **Index**: 004
 
 <!-- Status -->
-- **Status**: DRAFTING
+- **Status**: ACCEPTED
 <!-- **DRAFTING**/**ACCEPTED**/**REJECTED** -->
 
 <!-- Short description for the feature -->
--- a/docs/proposals/004-orchestration-refresh-certs.md
+++ b/docs/proposals/004-orchestration-refresh-certs.md
@@ -1,0 +1,266 @@
+<!--
+To start a new proposal, create a copy of this template on this directory and
+fill out the sections below.
+-->
+
+# Proposal information
+
+<!-- Index number -->
+- **Index**: 004
+
+<!-- Status -->
+- **Status**: DRAFTING
+<!-- **DRAFTING**/**ACCEPTED**/**REJECTED** -->
+
+<!-- Short description for the feature -->
+- **Name**: Cluster Orchestration - Certificate Refresh
+
+<!-- Owner name and github handle -->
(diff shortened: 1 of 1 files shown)


HomayoonAlimohammadi (approved):
Thanks a lot @mateoflorido! Something that I wonder is that do we have a retry mechanism on the single machine certs refresh reconciler? If that's the case, then you might need to consider the problem that we already have with the in-place upgrade reconcilers. 
In short, currently the single machine in-place upgrade reconciler retries by removing the `status` annotation on a machine without removing the upgrade instructions, which in a sense reverts everything back to a state that looks like a fresh upgrade command has been triggered. This can cause some confusions on the higher level which is the `MachineDeploymentReconciler`. 
There are workarounds for that tho, and I'm currently going for the one that I consider the best (adding a new annotation).

docs/proposals/004-orchestration-refresh-certs.md
```diff
@@ -0,0 +1,266 @@
...
+
+The status of the certificate refresh process will be shared via the
+`MachineDeployment` object in the same way as the `Cluster` controller.
+
+## ControlPlane Provider Changes
+<!--
+This section MUST mention any changes to the controlplane provider.
+-->
+
+A controller `CK8sCertificatesReconciler` will be added to the control plane
```
berkayoz, line 180:
```suggestion
A controller `ControlPlaneCertificatesReconciler` will be added to the control plane
```
Is this a copy paste error or intentional?

berkayoz (approved):
LGTM, great work left a small comment.

bschimke95 (commented):
LGTM, great proposal @mateoflorido

docs/proposals/004-orchestration-refresh-certs.md
mateoflorido: That was intentional, but I like your naming better. Thanks!

mateoflorido:
> Thanks a lot @mateoflorido! Something that I wonder is that do we have a retry mechanism on the single machine certs refresh reconciler? If that's the case, then you might need to consider the problem that we already have with the in-place upgrade reconcilers. In short, currently the single machine in-place upgrade reconciler retries by removing the `status` annotation on a machine without removing the upgrade instructions, which in a sense reverts everything back to a state that looks like a fresh upgrade command has been triggered. This can cause some confusions on the higher level which is the `MachineDeploymentReconciler`. There are workarounds for that tho, and I'm currently going for the one that I consider the best (adding a new annotation).

Great, we should consider this approach when we implement the proposal. Thanks, @HomayoonAlimohammadi!

bschimke95 (approved):
Let's change to accepted and merge the proposal.

PR #61, opened 2024-10-01
Use in-place upgrade release annotation if present

berkayoz:
This PR adjusts the ck8s config reconciler to use the snap install option from an in-place upgrade release annotation if it exists. This can be utilized by adding the release annotation under the `CK8sControlPlane` or `MachineDeployment` specs, which will be propagated down to the machine.

For control-plane (CK8sControlPlane)
`.spec.machineTemplate.metadata.annotations => Machine.annotations`

Adopted something similar to [propagation logic from KCP](https://github.com/kubernetes-sigs/cluster-api/blob/377b79c272a660064b97745bad85a6f77bc7a2d2/controlplane/kubeadm/internal/controllers/controller.go#L644-L751)

For workers (MachineDeployment)
`.spec.template.metadata.annotations => MachineSet.spec.template.metadata.annotations  => Machine.annotations`

### DIFF
--- a/bootstrap/controllers/ck8sconfig_controller.go
+++ b/bootstrap/controllers/ck8sconfig_controller.go
@@ -258,12 +258,15 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 		return err
 	}
 
+	snapInstallData := r.resolveInPlaceUpgradeRelease(machine)
+
 	input := cloudinit.JoinControlPlaneInput{
 		BaseUserData: cloudinit.BaseUserData{
 			BootCommands:        scope.Config.Spec.BootCommands,
 			PreRunCommands:      scope.Config.Spec.PreRunCommands,
 			PostRunCommands:     scope.Config.Spec.PostRunCommands,
 			KubernetesVersion:   scope.Config.Spec.Version,
+			SnapInstallData:     snapInstallData,
 			ExtraFiles:          cloudinit.FilesFromAPI(files),
 			ConfigFileContents:  string(joinConfig),
 			MicroclusterAddress: scope.Config.Spec.ControlPlaneConfig.MicroclusterAddress,
@@ -337,12 +340,15 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
 		return err
 	}
 
+	snapInstallData := r.resolveInPlaceUpgradeRelease(machine)
+
 	input := cloudinit.JoinWorkerInput{
 		BaseUserData: cloudinit.BaseUserData{
 			BootCommands:        scope.Config.Spec.BootCommands,
 			PreRunCommands:      scope.Config.Spec.PreRunCommands,
 			PostRunCommands:     scope.Config.Spec.PostRunCommands,
 			KubernetesVersion:   scope.Config.Spec.Version,
+			SnapInstallData:     snapInstallData,
 			ExtraFiles:          cloudinit.FilesFromAPI(files),
 			ConfigFileContents:  string(joinConfig),
 			MicroclusterAddress: scope.Config.Spec.ControlPlaneConfig.MicroclusterAddress,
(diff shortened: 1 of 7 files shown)


bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -391,6 +397,39 @@ func (r *CK8sConfigReconciler) resolveFiles(ctx context.Context, cfg *bootstrapv
 	return collected, nil
 }
 
+func (r *CK8sConfigReconciler) resolveInPlaceUpgradeRelease(machine *clusterv1.Machine) cloudinit.SnapInstallData {
+	mAnnotations := machine.GetAnnotations()
+
+	if mAnnotations != nil {
+		val, ok := mAnnotations[bootstrapv1.InPlaceUpgradeReleaseAnnotation]
```
eaudetcobello, line 404: You could invert the if statement here to remove a level of indentation, check if mAnnotations == nil and early return.

eaudetcobello (approved):
Really nice work, left one comment. Thanks!

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -685,6 +689,34 @@ func (r *CK8sControlPlaneReconciler) reconcileControlPlaneConditions(ctx context
...
+
+		patchHelper, err := patch.NewHelper(m, r.Client)
+		if err != nil {
+			return err
+		}
+
+		// Set annotations
+		// Add the annotations from the MachineTemplate.
+		for k, v := range kcp.Spec.MachineTemplate.ObjectMeta.Annotations {
+			m.Annotations[k] = v
```
HomayoonAlimohammadi, line 708: Is the possible that the `Annotations` field is nil?

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -685,6 +689,34 @@ func (r *CK8sControlPlaneReconciler) reconcileControlPlaneConditions(ctx context
...
+	for machineName := range controlPlane.Machines {
+		m := controlPlane.Machines[machineName]
+		// If the machine is already being deleted, we don't need to update it.
+		if !m.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		patchHelper, err := patch.NewHelper(m, r.Client)
+		if err != nil {
+			return err
```
HomayoonAlimohammadi, line 702: Maybe we can wrap this?

pkg/cloudinit/common.go
```diff
@@ -9,9 +9,26 @@ import (
 	"k8s.io/apimachinery/pkg/util/version"
 )
 
+type InstallOption string
+
+const (
+	InstallOptionChannel   InstallOption = "track"
```
HomayoonAlimohammadi, line 15: I'm a bit confused about this part. I can see `channel` instead of `track` and `localPath` instead of `local-path`.
https://github.com/canonical/cluster-api-k8s/blob/main/pkg/ck8s/workload_cluster.go#L279-L288

If these are the same, maybe we can centralize them?

pkg/cloudinit/scripts/install.sh
```diff
@@ -2,7 +2,19 @@
 
 ## Assumptions:
 ## - /capi/etc/snap-track contains the snap track that matches the installed Kubernetes version, e.g. "v1.30.1" -> "1.30-classic/stable"
```
HomayoonAlimohammadi, line 4: I think if the `track` needs to be replaced with `channel` according to the comments above, this one should be changes as well.

pkg/cloudinit/scripts/install.sh
```diff
@@ -2,7 +2,19 @@
...
+## - /capi/etc/snap-revision contains the snap revision to be installed, e.g. 123
+## - /capi/etc/snap-local-path contains the path to the local snap file to be installed, e.g. /path/to/k8s.snap
```
HomayoonAlimohammadi, line 6: And also we have `local-path` vs `localPath`

HomayoonAlimohammadi (commented):
Amazing work @berkayoz! Left some minor comments.

pkg/cloudinit/common.go
berkayoz: I've updated the `track` mentions to `channel`. For `localPath` vs `local-path`, I've kept `local-path` to fit the separate with `-` file naming convention we had. I'll adjust the `k8s-snap-api` to fit this form with another (cleanup) card I've this pulse.

HomayoonAlimohammadi (approved):
Great work and thanks a lot @berkayoz! LGTM

PR #62, opened 2024-10-01
Set SnapInstallData from CK8sConfigSpec

eaudetcobello:


HomayoonAlimohammadi (commented):
I feel like this needs to be rebased, right? https://github.com/canonical/cluster-api-k8s/pull/61

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -401,37 +413,57 @@ func (r *CK8sConfigReconciler) resolveInPlaceUpgradeRelease(machine *clusterv1.M
...
-			}
-		default:
-			r.Log.Info("Unknown in-place upgrade release option, ignoring", "option", optionKv[0])
 		}
 	}
 
 	return cloudinit.SnapInstallData{}
 }
 
+func (r *CK8sConfigReconciler) setSnapInstallDataFromSpec(spec bootstrapv1.CK8sConfigSpec) cloudinit.SnapInstallData {
```
bschimke95, line 445: nit: You don't set anything here. Let's call it `getSnapInstallDataFromSpec`

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -699,12 +699,7 @@ func (r *CK8sControlPlaneReconciler) syncMachines(ctx context.Context, kcp *cont
 
 		patchHelper, err := patch.NewHelper(m, r.Client)
 		if err != nil {
-			return fmt.Errorf("failed to create patch helper for machine: %w", err)
-		}
-
-		// Create a new map if machine has no annotations.
-		if m.Annotations == nil {
-			m.Annotations = map[string]string{}
+			return err
```
bschimke95, line 702: This will panic if no annotation is set. 

a) Use `m.GetAnnotations()` to avoid that
b) Wrap the error you return here

bootstrap/api/v1beta2/ck8sconfig_types.go
```diff
@@ -51,6 +51,20 @@ type CK8sConfigSpec struct {
 	// +optional
 	AirGapped bool `json:"airGapped,omitempty"`
 
+	// Track is the channel to use for the snap install.
+	// +optional
+	Track string `json:"track,omitempty"`
```
berkayoz, line 56:
```suggestion
	Channel string `json:"channel,omitempty"`
```

Let's rename to channel, track is just a [part of the channel.](https://snapcraft.io/docs/channels)

bschimke95 (commented):
Did a first pass @eaudetcobello 
looking good.

pkg/cloudinit/scripts/install.sh
```diff
@@ -1,13 +1,13 @@
...
+## - /capi/etc/snap-track contains the snap track that matches the installed Kubernetes version, e.g. "v1.30.1" -> "1.30-classic/stable"
 ## - /capi/etc/snap-revision contains the snap revision to be installed, e.g. 123
 ## - /capi/etc/snap-local-path contains the path to the local snap file to be installed, e.g. /path/to/k8s.snap
 
-if [ -f "/capi/etc/snap-channel" ]; then
-  snap_channel="$(cat /capi/etc/snap-channel)"
-  snap install k8s --classic --channel "${snap_channel}"
+if [ -f "/capi/etc/snap-track" ]; then
+  snap_track="$(cat /capi/etc/snap-track)"
+  snap install k8s --classic --channel "${snap_track}"
```
berkayoz, line 10: This file should not be changed, seems like a rebase issue.

controlplane/controllers/ck8scontrolplane_controller.go
berkayoz, line 702: Is this also a rebase issue?

berkayoz (commented):
Great work, left some comments.

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -401,37 +413,57 @@ func (r *CK8sConfigReconciler) resolveInPlaceUpgradeRelease(machine *clusterv1.M
 	mAnnotations := machine.GetAnnotations()
 
 	if mAnnotations != nil {
-		return cloudinit.SnapInstallData{}
```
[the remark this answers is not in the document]
eaudetcobello: Definitely, thanks for catching this!

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -258,7 +258,13 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 		return err
 	}
 
-	snapInstallData := r.resolveInPlaceUpgradeRelease(machine)
+	snapInstallData := r.setSnapInstallDataFromSpec(scope.Config.Spec)
+
+	inPlaceInstallData := r.resolveInPlaceUpgradeRelease(machine)
+
+	if inPlaceInstallData != (cloudinit.SnapInstallData{}) {
```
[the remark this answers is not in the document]
eaudetcobello: Right, resolveInPlaceUpgrade returns an "empty" struct. I'll change it to direct field comparison so that devs. don't need to know the default initialization value of SnapInstallData fields.

bootstrap/controllers/ck8sconfig_controller.go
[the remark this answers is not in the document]
berkayoz: Ah it is a bug, seems like I've missed the condition while inverting it.

bootstrap/api/v1beta2/ck8sconfig_types.go
```diff
@@ -66,6 +66,20 @@ type CK8sConfigSpec struct {
 	// +optional
 	SnapstoreProxyID string `json:"snapstoreProxyId,omitempty"`
 
+	// Channel is the channel to use for the snap install.
+	// +optional
+	Channel string `json:"channel,omitempty"`
+
+	// Revision is the revision to use for the snap install.
+	// If Channel is set, this will be ignored.
```
HomayoonAlimohammadi, line 74: Not sure if what I'm about to say is the best way to go, but I would prefer to have a validation that checks you've only entered one of these. Otherwise throws and error or warns you that hey things are conflicting. WDYT?

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -258,7 +258,13 @@ func (r *CK8sConfigReconciler) joinControlplane(ctx context.Context, scope *Scop
 		return err
 	}
 
-	snapInstallData := r.resolveInPlaceUpgradeRelease(machine)
+	snapInstallData := r.getSnapInstallDataFromSpec(scope.Config.Spec)
```
HomayoonAlimohammadi, line 261: Seems like `Config` is a pointer and can be nil, maybe we can check for that to prevent panics? I applicable, please have look at other occurrences as well.

HomayoonAlimohammadi (commented):
Thanks a lot @eaudetcobello! Great work! Almost LGTM. Just some minor comments.

bschimke95 (commented):
LGTM, nothing to add to Hues comments

bootstrap/controllers/ck8sconfig_controller.go
eaudetcobello: Eh, we create a Config struct and assign it at the start of the Reconcile function, so it's never nil. Also this would require change all over the codebase so it's out of scope for this PR.

HomayoonAlimohammadi (approved):
LGTM, Thanks a lot @eaudetcobello!

PR #63, opened 2024-10-02
Add machine deployment reconciler

HomayoonAlimohammadi:
### Summary
This PR adds the `MachineDeploymentReconciler` according to the recent proposal. This controller is responsible for watching for `upgrade-to` annotation on `MachineDeployment` object and then handing down the annotation to its owned `Machine`s and wait for them to finish upgrading.

### DIFF
--- a/bootstrap/api/v1beta2/inplace_upgrade_consts.go
+++ b/bootstrap/api/v1beta2/inplace_upgrade_consts.go
@@ -18,5 +18,5 @@ const (
 	InPlaceUpgradeInProgressEvent = "InPlaceUpgradeInProgress"
 	InPlaceUpgradeDoneEvent       = "InPlaceUpgradeDone"
 	InPlaceUpgradeFailedEvent     = "InPlaceUpgradeFailed"
-	InPlaceUpgradeCanceledEvent   = "InPlaceUpgradeCanceled"
+	InPlaceUpgradeCancelledEvent  = "InPlaceUpgradeCancelled"
 )
--- a/bootstrap/api/v1beta2/inplace_upgrade_consts.go
+++ b/bootstrap/api/v1beta2/inplace_upgrade_consts.go
@@ -1,10 +1,11 @@
 package v1beta2
 
 const (
-	InPlaceUpgradeToAnnotation       = "v1beta2.k8sd.io/in-place-upgrade-to"
-	InPlaceUpgradeStatusAnnotation   = "v1beta2.k8sd.io/in-place-upgrade-status"
-	InPlaceUpgradeReleaseAnnotation  = "v1beta2.k8sd.io/in-place-upgrade-release"
-	InPlaceUpgradeChangeIDAnnotation = "v1beta2.k8sd.io/in-place-upgrade-change-id"
+	InPlaceUpgradeToAnnotation                  = "v1beta2.k8sd.io/in-place-upgrade-to"
+	InPlaceUpgradeStatusAnnotation              = "v1beta2.k8sd.io/in-place-upgrade-status"
+	InPlaceUpgradeReleaseAnnotation             = "v1beta2.k8sd.io/in-place-upgrade-release"
+	InPlaceUpgradeChangeIDAnnotation            = "v1beta2.k8sd.io/in-place-upgrade-change-id"
+	InPlaceUpgradeLastFailedAttemptAtAnnotation = "v1beta2.k8sd.io/in-place-upgrade-last-failed-attempt-at"
 )
 
 const (
@@ -17,4 +18,5 @@ const (
 	InPlaceUpgradeInProgressEvent = "InPlaceUpgradeInProgress"
 	InPlaceUpgradeDoneEvent       = "InPlaceUpgradeDone"
 	InPlaceUpgradeFailedEvent     = "InPlaceUpgradeFailed"
+	InPlaceUpgradeCanceledEvent   = "InPlaceUpgradeCanceled"
 )
--- a/bootstrap/controllers/machine_deployment_controller.go
+++ b/bootstrap/controllers/machine_deployment_controller.go
@@ -80,24 +80,24 @@ func (r *MachineDeploymentReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	traceID := trace.NewID()
 	log := r.Log.WithValues("machine_deployment", req.NamespacedName, "trace_id", traceID)
-	log.Info("Reconciliation started...")
+	log.V(1).Info("Reconciliation started...")
 
 	machineDeployment := &clusterv1.MachineDeployment{}
 	if err := r.Get(ctx, req.NamespacedName, machineDeployment); err != nil {
 		if apierrors.IsNotFound(err) {
-			log.Info("MachineDeployment resource not found. Ignoring since object must be deleted.")
+			log.V(1).Info("MachineDeployment resource not found. Ignoring since object must be deleted.")
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, fmt.Errorf("failed to get MachineDeployment: %w", err)
 	}
 
-	if machineDeployment.Annotations[bootstrapv1.InPlaceUpgradeToAnnotation] == "" {
-		log.Info("MachineDeployment has no upgrade instructions, skipping reconciliation")
+	if r.getUpgradeInstructions(machineDeployment) == "" {
+		log.V(1).Info("MachineDeployment has no upgrade instructions, skipping reconciliation")
 		return ctrl.Result{}, nil
 	}
 
 	if !machineDeployment.DeletionTimestamp.IsZero() {
-		log.Info("MachineDeployment is being deleted, skipping reconciliation")
+		log.V(1).Info("MachineDeployment is being deleted, skipping reconciliation")
 		return ctrl.Result{}, nil
 	}
 
@@ -115,13 +115,13 @@ func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	var upgradedMachines int
 	for _, m := range ownedMachines {
 		if r.isMachineUpgraded(scope, m) {
-			log.Info("Machine is already upgraded", "machine", m.Name)
+			log.V(1).Info("Machine is already upgraded", "machine", m.Name)
 			upgradedMachines++
 			continue
 		}
 
 		if !m.DeletionTimestamp.IsZero() {
-			log.Info("Machine is being deleted, requeuing...", "machine", m.Name)
+			log.V(1).Info("Machine is being deleted, requeuing...", "machine", m.Name)
 			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 		}
 
@@ -135,7 +135,7 @@ func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 
 		if r.isMachineUpgrading(m) {
-			log.Info("Machine is upgrading, requeuing...", "machine", m.Name)
+			log.V(1).Info("Machine is upgrading, requeuing...", "machine", m.Name)
 			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 		}
 
@@ -144,7 +144,7 @@ func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, fmt.Errorf("failed to mark machine to upgrade: %w", err)
 		}
 
-		log.Info("Machine marked for upgrade", "machine", m.Name)
+		log.V(1).Info("Machine marked for upgrade", "machine", m.Name)
 
 		if err := r.markUpgradeInProgress(ctx, scope, m); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to mark upgrade as in-progress: %w", err)
@@ -158,7 +158,7 @@ func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, fmt.Errorf("failed to mark upgrade as done: %w", err)
 		}
 
-		log.Info("All machines are upgraded")
+		log.V(1).Info("All machines are upgraded")
 		return ctrl.Result{}, nil
 	}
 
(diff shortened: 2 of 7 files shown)


bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;machines/status,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinesets;machinesets/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinedeployments;machinedeployments/status,verbs=get;list;watch;create;update;patch;delete
+
+// Reconcile handles the reconciliation of a MachineDeployment object.
+func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// NOTE(Hue): Generate a tarce ID for this reconciliation to track logs better.
```
bschimke95, line 81: nit: typo *trace

bschimke95: tbh I would just remove this comment

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to get owned machines: %w", err)
+	}
+
+	scope, err := r.createScope(machineDeployment, ownedMachines)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create scope: %w", err)
+	}
+
+	if !r.hasUpgradeInstructions(scope) {
+		log.Info("MachineDeployment has no upgrade instructions, removing upgrade-to annotation from machines and stopping reconciliation")
```
bschimke95, line 111: This would silently remove any manual upgrade from a single machine. right? Is this by design and we are fine with that?

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+	var upgradedMachines int
+	for _, m := range ownedMachines {
+		if r.isMachineUpgraded(scope, m) {
+			log.Info("Machine is already upgraded", "machine", m.Name)
+			upgradedMachines++
+			continue
+		}
+
+		if !m.DeletionTimestamp.IsZero() {
+			if !m.DeletionTimestamp.IsZero() {
```
bschimke95, line 128: it is always good to double check but executing the same if statement twice in a nested block is probably not necessary ;)

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+				log.Info("Machine is being deleted", "machine", m.Name)
+			}
+			continue
+		}
+
+		if r.isMachineUpradeFailed(m) {
```
bschimke95, line 134: typo: Upgrade

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+		log.Info("All machines are upgraded")
+		return ctrl.Result{}, nil
+	}
+
+	return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+}
+
+// markUpgradeInProgress marks the MachineDeployment as in-place upgrade in-progress.
+func (r *MachineDeploymentReconciler) markUpgradeInProgress(ctx context.Context, scope *MachineDeploymentUpgradeScope) error {
+	ant := scope.MachineDeployment.Annotations
```
bschimke95, line 174: please just use the full word. There is no need for a reader to keep this (IMHO non-obvious) abbreviation in his mental model.

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+
+	if err := scope.PatchHelper.Patch(ctx, scope.MachineDeployment); err != nil {
+		return fmt.Errorf("failed to patch: %w", err)
+	}
+
+	r.recorder.Eventf(
+		scope.MachineDeployment,
+		corev1.EventTypeNormal,
+		bootstrapv1.InPlaceUpgradeDoneEvent,
+		"MachineDeployment %s marked as in-place upgrade done",
```
bschimke95, line 220: This is the user-facing string, right? 
Let's have something more straight forward then:
`MachineDeployment in-place upgrade done`. 
No need to tell that we "marked" a resource, the user cares about the outcome. (same above)

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+
+	if err := scope.PatchHelper.Patch(ctx, scope.MachineDeployment); err != nil {
+		return fmt.Errorf("failed to patch: %w", err)
+	}
+
+	r.recorder.Eventf(
+		scope.MachineDeployment,
+		corev1.EventTypeWarning,
+		bootstrapv1.InPlaceUpgradeFailedEvent,
+		"MachineDeployment %q marked as in-place upgrade failed. Upgrade failed for machine %q",
```
bschimke95, line 246: We could include the error message here. Would this make sense?

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,437 @@
...
+
+	var ms clusterv1.MachineSet
+	// NOTE(Hue): The nosec is due to a false positive: https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop
+	for _, _ms := range msList.Items { // #nosec G601
+		if util.IsOwnedByObject(&_ms, md) {
+			ms = _ms
+		}
+	}
+
+	ownedMachinesM, err := r.machineGetter.GetMachinesForCluster(ctx, client.ObjectKeyFromObject(cluster), collections.OwnedMachines(&ms))
```
bschimke95, line 316: what does the M stand for?

bootstrap/controllers/upgrade_controller.go
```diff
@@ -287,6 +291,13 @@ func (r *InPlaceUpgradeReconciler) handleUpgradeDone(ctx context.Context, scope
 func (r *InPlaceUpgradeReconciler) handleUpgradeFailed(ctx context.Context, scope *UpgradeScope) (reconcile.Result, error) {
 	mAnnotations := scope.Machine.GetAnnotations()
 
+	// NOTE(Hue): We don't remove the `LastFailedAttemptAt` annotation here
+	// because we want to know if the upgrade failed at some point in the `MachineDeploymentReconciler`.
+	// This function triggers a retry by removing the `Status` and `ChangeID` annotations,
+	// but the `LastFailedAttemptAt` lets us descriminiate between a retry and a fresh upgrade.
+	// Overall, we don't remove the `LastFailedAttemptAt` annotation in the `InPlaceUpgradeReconciler`.
+	// That's the responsibility of the `MachineDeploymentReconciler`.
```
bschimke95, line 299: Mote: I'm not super happy about this solution. It causes an implicit dependency between two controllers but I don't really have an alternative solution so I'm fine with it for now.

pkg/trace/trace.go
```diff
@@ -0,0 +1,17 @@
+package trace
+
+import (
+	"crypto/rand"
+	"math/big"
+)
+
+const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+func NewID() string {
```
bschimke95, line 10: Add a doc string please

bschimke95 (commented):
Great work @HomayoonAlimohammadi
added a couple of comments

bootstrap/controllers/machine_deployment_controller.go
HomayoonAlimohammadi: You know I can't stop being suuuper careful :smile:.
JK. Thanks for the catch!

bootstrap/controllers/machine_deployment_controller.go
HomayoonAlimohammadi: It stands for map since the `collections.Machines`  is basically a `map[string]*clusterv1.Machine`. I can change it if you don't like it.

bootstrap/controllers/upgrade_controller.go
HomayoonAlimohammadi: I understand. This coupling is indeed not desirable. But given the other solutions that we discussed, I feel like this is an acceptable workaround so far.

bootstrap/controllers/machine_deployment_controller.go
HomayoonAlimohammadi: This is a really good idea, but I think it's not easily feasible right now. The single machine in-place upgrade reconciler (which has access to the original error) publishes an event for that but does not include it anywhere else: https://github.com/canonical/cluster-api-k8s/blob/main/bootstrap/controllers/upgrade_controller.go#L189-L204

bootstrap/controllers/machine_deployment_controller.go
HomayoonAlimohammadi: To be fair this was not mentioned in the proposal. But I did this intentionally to prevent unnecessary confusion and complexity later on (maybe I just added more confusion?). The rationale behind is that I think the users are supposed to use the `MachineDeployment` annotation and not the single machine ones. I can remove this if you think it's not the correct move.

bootstrap/api/v1beta2/inplace_upgrade_consts.go
```diff
@@ -17,4 +18,5 @@ const (
 	InPlaceUpgradeInProgressEvent = "InPlaceUpgradeInProgress"
 	InPlaceUpgradeDoneEvent       = "InPlaceUpgradeDone"
 	InPlaceUpgradeFailedEvent     = "InPlaceUpgradeFailed"
+	InPlaceUpgradeCanceledEvent   = "InPlaceUpgradeCanceled"
```
berkayoz, line 21:
```suggestion
	InPlaceUpgradeCancelledEvent   = "InPlaceUpgradeCancelled"
```

bootstrap/controllers/machine_deployment_controller.go
berkayoz: Users should be able to perform single machine operations, the orchestration should be optional. We would've directly implemented the upgrade logic in the higher level reconcilers such as `MachineDeployment` reconciler if the intention was to not perform single machine operations.

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,433 @@
...
+	annotations := scope.MachineDeployment.Annotations
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+
+	// clean up
+	delete(annotations, bootstrapv1.InPlaceUpgradeToAnnotation)
+
+	annotations[bootstrapv1.InPlaceUpgradeStatusAnnotation] = bootstrapv1.InPlaceUpgradeDoneStatus
+	annotations[bootstrapv1.InPlaceUpgradeReleaseAnnotation] = scope.UpgradeTo
```
berkayoz, line 207: We should also add this under `MachineDeployment.spec.template.metadata.annotations` which would propagate to `MachineSet.spec.template.metadata.annotations` and lastly propagate to `Machine.annotations`. This is important for scale up/rolling upgrade compatability. 

Also we should remove this annotation from both sections once an upgrade is requested, and add it after completion.

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,433 @@
...
+
+// createScope creates a new MachineDeploymentUpgradeScope.
+func (r *MachineDeploymentReconciler) createScope(md *clusterv1.MachineDeployment, ownedMachines []*clusterv1.Machine) (*MachineDeploymentUpgradeScope, error) {
+	patchHelper, err := patch.NewHelper(md, r.Client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new patch helper: %w", err)
+	}
+
+	// NOTE(Hue): The reason we are checking the `Release` annotation as well is that we want to make sure
+	// we upgrade the new machines that joined after the initial upgrade process.
```
berkayoz, line 257: I'm not sure what's the intention here, the `CK8sConfigReconciler` checks for the release annotation and overrides the install already. I believe just propagating the release annotation should be enough.

berkayoz (commented):
Great work! Did an initial pass and left some comments.

bootstrap/controllers/machine_deployment_controller.go
HomayoonAlimohammadi: Oh I was not aware of this: https://github.com/canonical/cluster-api-k8s/blob/main/bootstrap/controllers/ck8sconfig_controller.go#L400
Thanks for pointing this out.

bootstrap/controllers/machine_deployment_controller.go
HomayoonAlimohammadi: Right, will make the changes. But as a question:
For the scale up we will trigger the reconciliation and add the `upgrade-to` to the machines (manually in this reconciler), this should be enough, right? I also tested scale ups and everything works as expected (new machines get annotated and upgraded).
For rolling upgrade, I'm not sure what we should do here... 
Let's say we previously had an in-place upgrade by annotating the `MachineDeployment`. What should happen when the user triggers a rolling upgrade (by changing `version`) while also keeping the in-place upgrade annotation?

HomayoonAlimohammadi: Worth mentioning that while the current implementation _does_ cover scaling up, it's probably not standard (in terms of k8s and how annotation propagation works) nor aligned with how the `Ck8sConfigReconciler` works (uses `release` to overwrite snap install instructions)
So I'll change this. Thanks for pointing this out.

bootstrap/controllers/machine_deployment_controller.go
```diff
@@ -0,0 +1,414 @@
...
+	slices.SortStableFunc(ownedMachines, func(m1, m2 *clusterv1.Machine) int {
+		switch {
+		case m1.UID < m2.UID:
+			return -1
+		case m1.UID == m2.UID:
+			return 0
+		default:
+			return 1
+		}
+	})
```
eaudetcobello, line 346: Are there any guarantees that UID's increase monotonically? Looks like that's a prerequisite for this sort to work.

eaudetcobello (changes requested):
This is really great work overall. I apologize for being a bit insistent about the sorting logic—it's just the one part of the PR that doesn't sit quite right with me.

bschimke95 (approved):
LGTM, great work!

HomayoonAlimohammadi:
Worth noting that the added commits after the LGTM are just because of a rebase (with no conflicting changes).

PR #64, opened 2024-10-02
Support Snapstore Proxy Configuration

eaudetcobello:
Adds support for configuring a snap store proxy through the manifest.

All the configuration and wiring happens in NewBaseCloudConfig so that we don't need to reimplement the same code for each type of join/init (worker, cp).

**Important**: One thing that might be worth noting is that, if it's needed, `configure-snapstore-proxy.sh` will be run *before* PreRunCommands. See the updated tests.


pkg/cloudinit/controlplane_init.go
```diff
@@ -53,6 +53,9 @@ func NewInitControlPlane(input InitControlPlaneInput) (CloudConfig, error) {
 
 	// run commands
 	config.RunCommands = append(config.RunCommands, input.PreRunCommands...)
+
+	config.RunCommands = append(config.RunCommands, "/capi/scripts/configure-snapstore-proxy.sh")
```
berkayoz, line 57: Can we check if proxy is configured properly in the spec to before adding this script to `RunCommands`?

pkg/cloudinit/scripts/configure-snapstore-proxy.sh
```diff
@@ -0,0 +1,35 @@
...
+#   - snapd is installed
+#   - /capi/etc/snapstore-proxy-scheme contains the snapstore scheme
+#   - /capi/etc/snapstore-proxy-domain contains the snapstore domain
+#   - /capi/etc/snapstore-proxy-id contains the snapstore id
+
+SNAPSTORE_PROXY_SCHEME=$(cat /capi/etc/snapstore-proxy-scheme)
+SNAPSTORE_PROXY_DOMAIN=$(cat /capi/etc/snapstore-proxy-domain)
+SNAPSTORE_PROXY_ID=$(cat /capi/etc/snapstore-proxy-id)
+
+if [ -z "${SNAPSTORE_PROXY_SCHEME}" ] || [ -z "${SNAPSTORE_PROXY_DOMAIN}" ] || [ -z "${SNAPSTORE_PROXY_ID}" ]; then
```
berkayoz, line 13: Any reason to not check for file existence? `cat` will fail on a non existent path, how does this interact with `-z` ?

pkg/cloudinit/scripts/configure-snapstore-proxy.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Missing snapstore proxy configuration"
+  exit 1
+fi
+
+if ! type -P curl; then
+  while ! snap install curl; do
```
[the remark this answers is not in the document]
berkayoz: +1, we should not have infinite retries. We can fail the while script after certain number of retries or timeout.

pkg/cloudinit/common.go
```diff
@@ -54,6 +60,11 @@ func NewBaseCloudConfig(data BaseUserData) (CloudConfig, error) {
 			Owner:       "root:root",
 		})
 	}
+
+	// snapstore proxy config files
+	snapStoreConfigFiles := getSnapstoreProxyConfigFiles(data.SnapstoreProxyScheme, data.SnapstoreProxyDomain, data.SnapstoreProxyID)
+	config.WriteFiles = append(config.WriteFiles, snapStoreConfigFiles...)
```
berkayoz, line 66: Should we check `if snapStoreConfigFiles != nil` first?

berkayoz (commented):
Great work, did an initial pass

pkg/cloudinit/common.go
```diff
@@ -101,3 +112,37 @@ func NewBaseCloudConfig(data BaseUserData) (CloudConfig, error) {
 func makeMicroclusterAddress(address string, port int) string {
 	return net.JoinHostPort(address, strconv.Itoa(port))
 }
+
+func getSnapstoreProxyConfigFiles(snapstoreProxyScheme, snapstoreProxyDomain, snapstoreProxyID string) []File {
```
bschimke95, line 116: You conditionally write the snapstore config files (scheme, domain, id files) but always write and execute the `configure-snapstore-config.sh`. 
In the script, you fail/exit if any of the configs are not set. 

This breaks cloud-init in the default configuration.

bschimke95 (commented):
Added an initial comment.

Please also properly document/describe your PR again.

pkg/cloudinit/scripts/configure-snapstore-proxy.sh
```diff
@@ -0,0 +1,35 @@
...
+#   - /capi/etc/snapstore-proxy-domain contains the snapstore domain
+#   - /capi/etc/snapstore-proxy-id contains the snapstore id
+
+SNAPSTORE_PROXY_SCHEME=$(cat /capi/etc/snapstore-proxy-scheme)
+SNAPSTORE_PROXY_DOMAIN=$(cat /capi/etc/snapstore-proxy-domain)
+SNAPSTORE_PROXY_ID=$(cat /capi/etc/snapstore-proxy-id)
+
+if [ -z "${SNAPSTORE_PROXY_SCHEME}" ] || [ -z "${SNAPSTORE_PROXY_DOMAIN}" ] || [ -z "${SNAPSTORE_PROXY_ID}" ]; then
+  echo "Missing snapstore proxy configuration"
+  exit 1
```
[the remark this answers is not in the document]
eaudetcobello: No, the commands continue.

pkg/cloudinit/scripts/configure-snapstore-proxy.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Failed to ACK store assertions, will retry"
+  sleep 5
+done
+
+while ! snap set core proxy.store="${SNAPSTORE_PROXY_ID}"; do
+  echo "Failed to configure snapd with store ID, will retry"
+  sleep 5
+done
+
+systemctl restart snapd
```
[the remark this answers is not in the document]
eaudetcobello: systemctl restart is blocking

pkg/cloudinit/scripts/configure-snapstore-proxy.sh
eaudetcobello: If one of the files don't exist then we'll never reach the if. This is because of `set -e`. I'll think of a better solution.

pkg/cloudinit/common.go
eaudetcobello: It doesn't break, because cloud-init doesn't stop when one of it's runcmds fail. Regardless, I'll make this more robust.

HomayoonAlimohammadi (approved):
Great job @eaudetcobello! Thanks a lot. LGTM.

addyess:
This doesn't block the work, but can i change my snapstore proxy after install?  

say it starts at foo.svc:1234 and i move the server to bar.svc:1234

are we prepared for that or do we need another card to track that?

eaudetcobello:
> This doesn't block the work, but can i change my snapstore proxy after install?
> 
> say it starts at foo.svc:1234 and i move the server to bar.svc:1234
> 
> are we prepared for that or do we need another card to track that?

You would have to apply the manifest with the new address and create a new control plane. We can create a card to reconcile proxy configuration if you want.

PR #66, opened 2024-10-08
Support user-supplied bootstrap configuration

eaudetcobello:
Implements first-party support for supplying a bootstrap configuration (before this, it was done by overriding install.sh and adding a file under extraFiles). 

The BootstrapConfig type re-uses the existing FileSource type so that a secret reference can be used to seed the configuration just like we have for files, (e.g. contentFrom).


bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -433,18 +459,18 @@ func (r *CK8sConfigReconciler) resolveInPlaceUpgradeRelease(machine *clusterv1.M
 }
 
 // resolveSecretFileContent returns file content fetched from a referenced secret object.
-func (r *CK8sConfigReconciler) resolveSecretFileContent(ctx context.Context, ns string, source bootstrapv1.File) ([]byte, error) {
+func (r *CK8sConfigReconciler) resolveSecretFileContent(ctx context.Context, ns string, source bootstrapv1.FileSource) ([]byte, error) {
```
eaudetcobello, line 462: I changed the parameter type from File to FileSource because 1. We did not use any of the fields from File, except ContentFrom. 2. This makes the function reusable for any other types that have a FileSource.

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -564,6 +590,12 @@ func (r *CK8sConfigReconciler) handleClusterNotInitialized(ctx context.Context,
 		return ctrl.Result{}, err
 	}
 
+	userSuppliedBootstrapConfig, err := r.resolveUserBootstrapConfig(ctx, scope.Config)
+	if err != nil {
+		conditions.MarkFalse(scope.Config, bootstrapv1.DataSecretAvailableCondition, bootstrapv1.DataSecretGenerationFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
+		return ctrl.Result{}, err
+	}
+
```
eaudetcobello, line 598: resolveUserBootstrapConfig returns an empty string if there is no user-supplied configuration, we check this later in common.go.

pkg/cloudinit/common.go
```diff
@@ -78,14 +80,24 @@ func NewBaseCloudConfig(data BaseUserData) (CloudConfig, error) {
...
 	}
+
+	var configFileContents string
+	switch {
+	case data.BootstrapConfig != "":
+		// User-supplied bootstrap configuration from CK8sConfig object.
+		configFileContents = data.BootstrapConfig
+	default:
+		configFileContents = data.ConfigFileContents
+	}
```
eaudetcobello, line 91: user-supplied bootstrap configuration has precedence over our ConfigFileContents

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -376,6 +376,32 @@ func (r *CK8sConfigReconciler) joinWorker(ctx context.Context, scope *Scope) err
...
+	// User referenced a secret for the bootstrap configuration
+	if cfg.Spec.BootstrapConfig.ContentFrom == nil {
+		return "", nil
+	}
+
+	data, err := r.resolveSecretFileContent(ctx, cfg.Namespace, *cfg.Spec.BootstrapConfig.ContentFrom)
+	if err != nil {
+		return "", fmt.Errorf("failed to read bootstrap configuration from secret: %w", err)
```
bschimke95, line 399: nit: I'd add the secret name ref to the error message

pkg/cloudinit/common.go
```diff
@@ -35,6 +35,8 @@ type BaseUserData struct {
 	PreRunCommands []string
 	// PostRunCommands is a list of commands to run after k8s installation.
 	PostRunCommands []string
+	// BootstrapConfig boop bop
```
bschimke95, line 38: blub

pkg/cloudinit/controlplane_init_test.go
```diff
@@ -92,6 +95,31 @@ func TestNewInitControlPlane(t *testing.T) {
 	), "Some /capi/scripts files are missing")
 }
 
+func TestUserSuppliedBootstrapConfig(t *testing.T) {
+	g := NewWithT(t)
+
+	config, err := cloudinit.NewInitControlPlane(cloudinit.InitControlPlaneInput{
+		BaseUserData: cloudinit.BaseUserData{
+			KubernetesVersion:  "v1.30.0",
```
bschimke95, line 103: nit: let's make this 1.31. since we don't support lower versions.

templates/aws/cluster-template.yaml
```diff
@@ -51,6 +51,13 @@ spec:
       name: ${CLUSTER_NAME}-control-plane
   spec:
     nodeName: "{{ ds.meta_data.local_hostname }}"
+    bootstrapConfig:
```
bschimke95, line 54: this is probably to be removed before merging, right?

pkg/cloudinit/common.go
```diff
@@ -78,14 +80,24 @@ func NewBaseCloudConfig(data BaseUserData) (CloudConfig, error) {
 			Owner:       "root:root",
 		})
 	}
+
+	var configFileContents string
+	switch {
+	case data.BootstrapConfig != "":
```
bschimke95, line 86: why switch case and not a simple if block?

bschimke95 (commented):
Looks good, added some comments

pkg/cloudinit/common.go
eaudetcobello: 😂 Forgot this

pkg/cloudinit/common.go
eaudetcobello: Go linter barked at me once to change an if to a switch and i have been traumatized ever since

templates/aws/cluster-template.yaml
eaudetcobello: Yes

pkg/cloudinit/common.go
bschimke95: Ok, I played around with golangci yesterday and it was barking at me the other way around so you might be barked at again if we introduce that here.
But feel free to ignore it until then - I don't have a strong opinion (here and in general).

pkg/cloudinit/controlplane_init_test.go
eaudetcobello: I'll create a card to update all the strings to 1.31 instead of just this one.

bschimke95 (approved):
LGTM

PR #69, opened 2024-10-11
Improve UX around setting Proxy configurations in the spec

maci3jka:
This PR introduces `httpProxy`, `httpsProxy`, `noProxy` parameters that adds support for proxy settings for contained in k8s. Provides functionality mentioned in [k8s docs](https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/proxy/).
KU-1801


bootstrap/api/v1beta2/ck8sconfig_types.go
```diff
@@ -66,6 +66,18 @@ type CK8sConfigSpec struct {
...
+	// +optional
+	HTTPSProxy string `json:"httpsProxy,omitempty"`
+
+	// HTTPProxy is optional http proxy configuration
+	// +optional
+	HTTPProxy string `json:"httpProxy,omitempty"`
+
+	// NoProxy is optional no proxy configuration
+	// +optional
+	NoProxy string `json:"noProxy,omitempty"`
```
berkayoz, line 79: I'm not too happy with the naming here as these are only getting set for containerd. We should make it explicit like `ContainerdHTTPProxy`  if this is what we want.

berkayoz (commented):
LGTM overall, I have one nit/question.

@bschimke95 Should we follow the k8s-operator approach of adding these variables to `/etc/environment` instead? If not we should make the variables reflect this is only for containerd.

pkg/cloudinit/controlplane_join_test.go
```diff
@@ -82,6 +83,86 @@ func TestNewJoinControlPlane(t *testing.T) {
...
+			BootCommands:      []string{"bootcmd"},
+			PreRunCommands:    []string{"prerun1", "prerun2"},
+			PostRunCommands:   []string{"postrun1", "postrun2"},
+			ExtraFiles: []cloudinit.File{{
+				Path:        "/tmp/file",
+				Content:     "test file",
+				Permissions: "0400",
+				Owner:       "root:root",
+			}},
+			SnapstoreProxyScheme: "http",
```
eaudetcobello, line 101: Let's remove snapstore proxy fields from the tests of containerd proxy, to make it clear that they are not related

pkg/cloudinit/controlplane_join_test.go
```diff
@@ -82,6 +83,86 @@ func TestNewJoinControlPlane(t *testing.T) {
...
+		JoinToken: "test-token",
+	})
+
+	g.Expect(err).NotTo(HaveOccurred())
+
+	// Verify the boot commands.
+	g.Expect(config.BootCommands).To(Equal([]string{"bootcmd"}))
+
+	// Verify the run commands.
+	g.Expect(config.RunCommands).To(Equal([]string{
```
eaudetcobello, line 119: It's not necessary to re-test all these items. Only test the relevant ones since we have a test for all of them already. 

Instead, only test the items that would be affected by setting a containerd proxy . For example, test that the containerd proxy setup script is present

eaudetcobello (changes requested):
Looking good. Please apply my comments to the tests we added. We don't need to duplicate all the lines that are already tested by other tests, instead let's only check for the fields we know are relevant to containerd proxy configuration

pkg/cloudinit/controlplane_init_test.go
```diff
@@ -105,6 +107,103 @@ func TestNewInitControlPlane(t *testing.T) {
 	), "Some /capi/scripts files are missing")
 }
 
+func TestNewInitControlPlaneWithOptionalProxySettings(t *testing.T) {
+	g := NewWithT(t)
+	for _, tc := range []struct {
```
maci3jka, line 112: Is adding multiple test cases ok or too excessive?

berkayoz: Checking it in 2 cases should be enough. One to see the files are appended when proxy is configured, and maybe one to see the files are not created when proxy is not configured.

pkg/cloudinit/common.go
```diff
@@ -190,3 +201,32 @@ func getSnapstoreProxyConfigFiles(data BaseUserData) []File {
...
 }
+
+// getProxyConfigFiles returns the node proxy config files.
+// If the HTTPProxy or HTTPPSProxy is not set, it returns nil.
+// Nil indicates that no files are returned.
+func getProxyConfigFiles(data BaseUserData) []File {
+	if data.HTTPProxy == "" && data.HTTPSProxy == "" {
+		return nil
+	}
+	return []File{
```
berkayoz, line 212: Let's check if the variable is provided to add the respective file to the list. For example
```
if data.HTTPProxy != "" {
   files = append(files, File{...})
}
```

bootstrap/config/crd/bases/bootstrap.cluster.x-k8s.io_ck8sconfigs.yaml
```diff
@@ -87,6 +84,9 @@ spec:
                     - secret
                     type: object
                 type: object
+              channel:
```
berkayoz, line 87: Why does this change here?
Did we forget running `make generate && make manifests` before?

berkayoz (commented):
Thank you, we're almost there some final comments

bootstrap/config/crd/bases/bootstrap.cluster.x-k8s.io_ck8sconfigs.yaml
maci3jka: looks like it `make generate && make manifests` created it like that for me

pkg/cloudinit/controlplane_init_test.go
maci3jka: I left 3 additonaly with one field to make sure that logic behaves as it sould

berkayoz (approved):
LGTM overall, one small nit before merging.

bschimke95 (approved):
LGTM

pkg/cloudinit/scripts/configure-proxy.sh
```diff
@@ -0,0 +1,31 @@
...
+  echo "https_proxy=${HTTPS_PROXY}" >> "${ENVIRONMENT_FILE}"
+  echo "HTTPS_PROXY=${HTTPS_PROXY}" >> "${ENVIRONMENT_FILE}"
+fi
+
+
+if [ -f ${NO_PROXY_FILE} ]; then
+  NO_PROXY=$(cat ${NO_PROXY_FILE})
+  echo "no_proxy=${NO_PROXY}" >> "${ENVIRONMENT_FILE}"
+  echo "NO_PROXY=${NO_PROXY}" >> "${ENVIRONMENT_FILE}"
+fi
```
eaudetcobello, line 31: Can you make it so there are only 1 line between the if's? Same thing between lines 9 and 13.

eaudetcobello (approved):

PR #71, opened 2024-10-15
Add CP orchestrated in-place upgrade controller

HomayoonAlimohammadi:
### Overview
This PR adds the **orchestrated in-place upgrade controller** to the control plane provider.
After annotating the `CK8sControlPlane` object with `inplace-upgrade-to` annotation, this controller will make sure that each control plane machine is upgraded to the specified version.
By design, this controller does not allow parallel upgrades to ensure that the control plane stays responsive (prevent breaking HA, quorum, etc.)

### Details
- A lock (which is just a glorified configMap) prevents multiple control plane nodes getting upgraded
- Some of the common functionality is moved to `pkg/upgrade/inplace`. We should use these functionalities for the `MachineDeploymentReconciler` as well (instead of separately implementing them, which is what we're doing right now).
    - Also the `MachineDeploymentReconciler` should get renamed to resemble its responsibilities more closely (e.g. `bootstrapControllers.OrchestratedInPlaceUpgradeController`) in the future.

### DIFF
--- a/controlplane/controllers/orchestrated_inplace_upgrade_controller.go
+++ b/controlplane/controllers/orchestrated_inplace_upgrade_controller.go
@@ -1,0 +1,301 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/collections"
+	"sigs.k8s.io/cluster-api/util/patch"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	bootstrapv1 "github.com/canonical/cluster-api-k8s/bootstrap/api/v1beta2"
+	controlplanev1 "github.com/canonical/cluster-api-k8s/controlplane/api/v1beta2"
+	"github.com/canonical/cluster-api-k8s/pkg/ck8s"
+	"github.com/canonical/cluster-api-k8s/pkg/trace"
+	"github.com/canonical/cluster-api-k8s/pkg/upgrade/inplace"
+)
+
+// OrchestratedInPlaceUpgradeController reconciles a CK8sControlPlane object and manages the in-place upgrades.
+type OrchestratedInPlaceUpgradeController struct {
+	scheme        *runtime.Scheme
+	recorder      record.EventRecorder
+	machineGetter inplace.MachineGetter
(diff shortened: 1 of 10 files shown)


controlplane/controllers/orchestrated_inplace_upgrade_controller.go
```diff
@@ -0,0 +1,301 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to create scope: %w", err)
+	}
+
+	lockedMachine, err := r.lock.IsLocked(ctx, scope.cluster)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to check if upgrade is locked: %w", err)
+	}
+
+	// Upgrade is locked and a machine is already upgrading
+	if lockedMachine != nil {
```
berkayoz, line 113: What happens on if upgrade is failed on the machine, do we wait/hold the lock until it succeds?

berkayoz (commented):
Amazing work! Can you explain in short form how the lock flow works, especially around failures and retries.

controlplane/controllers/orchestrated_inplace_upgrade_controller.go
HomayoonAlimohammadi: Thanks for pointing this out, I noticed that I had to move the `IsMachineUpgradeFailed` check a bit higher.

In short, yes. Lemme explain a bit further:

### Scenario 1: Not locked
- Let's say the upgrade process is not locked. We select a machine to upgrade (let's call it `M`) and lock the upgrade process while mentioning in the lock object that machine `M` is upgrading.

### Scenario 2: Locked and the machine is upgraded
- We see that the process is locked, so we get machine `M` and notice that it is upgraded (`release` annotation). We unlock and requeue the reconciliation.

### Scenario 3: Locked and the machine is not upgraded (still upgrading)
- We requeue the reconciliation.

### Scenario 4: A machine that we tried to upgrade before is not there any more (somehow deleted)
- The lock config map has the machine information and tries to obtain it but notices that it's not there, so the `IsLocked` won't return an `upgradingMachine`, the lock config map is deleted (`Unlock` is called) and we continue on to select a fresh machine to upgrade.

### Scenario 5: Locked and the machine does not have any upgrade instructions
- This is a weird situation (maybe the admin deleted the annotations accidentally), but _can_ happen nonetheless. We consider that a stale lock and Unlock and requeue.

### Scenario 6: Locked and the machine upgrade failed
- We mark the `CK8sControlPlane` object with `failed` status and publish the respective event and retry the process (basically the underlying `single machine in-place upgrade reconciler` is doing the retry and we just wait).

bschimke95 (commented):
Generally LGTM but please create some unittests for the `inplace` package and an integration test that verifies the happy path of this feature.

controlplane/controllers/orchestrated_inplace_upgrade_controller.go
bschimke95: Thanks for the list - this was very helpful for the review.

bschimke95 (approved):
LGTM, thanks for adding the integration test!

eaudetcobello (approved):
Fantastic work, thanks Hue

PR #73, opened 2024-10-16
Use new channel field instead of install script overwrite

eaudetcobello:
We now have a channel field we can use to specify the installation channel. This PR removes the install script overwrite and replaces it with the channel field.


bschimke95 (approved):
nice

PR #78, opened 2024-10-31
 Add unittest for `/pkg/upgrade`

maci3jka:
This pr introduces unit tests for `pkg/upgrade/inplace` package
 KU-1891

### DIFF
--- a/pkg/upgrade/inplace/inplace_test.go
+++ b/pkg/upgrade/inplace/inplace_test.go
@@ -1,0 +1,209 @@
+package inplace_test
+
+import (
+	"testing"
+	"time"
+
+	bootstrapv1 "github.com/canonical/cluster-api-k8s/bootstrap/api/v1beta2"
+	controlplanev1 "github.com/canonical/cluster-api-k8s/controlplane/api/v1beta2"
+	"github.com/canonical/cluster-api-k8s/pkg/upgrade/inplace"
+
+	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


pkg/upgrade/inplace/inplace_test.go
```diff
@@ -0,0 +1,209 @@
...
+	g := NewWithT(t)
+
+	for _, tc := range []struct {
+		name        string
+		annotations map[string]string
+		upgradeTo   string
+		isUpgraded  bool
+	}{
+		{
+			name: "InPlaceUpgradeReleaseAnnotationOnly",
```
HomayoonAlimohammadi, line 68: Maybe we can also add `InPlaceUpgradeToAnnotationOnly`

pkg/upgrade/inplace/inplace_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	for _, tc := range []struct {
+		name          string
+		annotations   map[string]string
+		releaseString string
+		isUpgraded    bool
+	}{
+		{
+			name:          "notUpgraded",
+			annotations:   map[string]string{bootstrapv1.InPlaceUpgradeReleaseAnnotation: "v1.30"},
```
HomayoonAlimohammadi, line 27: Let's change the v1.30 and v1.29 here, not that it's important in passing the test but only that it makes more sense because we expected the machine actual version be lower than the upgrade goal.

HomayoonAlimohammadi (approved):
Thanks a lot @maci3jka, LGTM! Just some minor comments.

PR #81, opened 2024-11-08
Refactor machine deployment reconciler

HomayoonAlimohammadi:
### Overview
This PR changes how the (the now renamed) `MachineDeploymentReconciler` is implemented.
List of changes:
- Rename to `OrchestratedInPlaceUpgrade` (there are two of these, one is in `./controlplane` and the other in `./bootstrap`)
- Use common logic from the `inplace` package (used in both machine deployment and control plane orchestrated upgrade controllers)


bootstrap/controllers/orchestrated_inplace_upgrade_controller.go
```diff
@@ -91,41 +87,36 @@ func (r *MachineDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to create scope: %w", err)
 	}
 
 	// Starting the upgrade process
 	var upgradedMachines int
-	for _, m := range ownedMachines {
-		if r.isMachineUpgraded(scope, m) {
+	for _, m := range scope.ownedMachines {
+		if inplace.IsUpgraded(m, scope.upgradeTo) {
```
eaudetcobello, line 108: Out of curiosity, why do we go from ownedMachines to scope.ownedMachines?

eaudetcobello (commented):
Nice cleanup! Thanks!

eaudetcobello (approved):

bootstrap/controllers/orchestrated_inplace_upgrade_controller.go
HomayoonAlimohammadi: Just so that there's a single reference/thing-we-can-use. Before, we declared an `ownedMachines` variable and passed it to the `createScope` function. IMHO that didn't really help with dependency injection and I believe we're good to let the `createScope` to also create the `ownedMachines` in itself. So now, we only have the `ownedMachines` through the `scope.owedMachines`.

PR #82, opened 2024-11-13
KU-2044

kubeages:
This adds the channel field in the MAAS template.


HomayoonAlimohammadi (approved):
Thanks a lot @kubeages! LGTM! Maybe we can also add these to other templates (e.g. AWS)?

eaudetcobello:
> Thanks a lot @kubeages! LGTM! Maybe we can also add these to other templates (e.g. AWS)?

We have it for AWS: https://github.com/canonical/cluster-api-k8s/blob/main/templates/aws/cluster-template.yaml#L55

PR #85, opened 2024-12-04
Fix default k8s snap channel

petrutlucian94:
This commit ensures that the "-classic/stable" channel is used if no snap channel, revision or path is provided.

### DIFF
--- a/pkg/cloudinit/common.go
+++ b/pkg/cloudinit/common.go
@@ -74,10 +74,9 @@ func NewBaseCloudConfig(data BaseUserData) (CloudConfig, error) {
 	snapInstall := data.SnapInstallData
 	// Default to k8s version if snap install option is not set or empty.
 	if snapInstall == nil || snapInstall.Option == "" {
-		// TODO: consider using "stable" once we'll have a stable channel.
 		snapInstall = &SnapInstallData{
 			Option: InstallOptionChannel,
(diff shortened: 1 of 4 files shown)


petrutlucian94:
> The default should always be stable, even if we don't have the tracks yet. We can adjust to use the 1.32 stable track which should be released soon.

I see, I've switched the default channel back to "stable".

pkg/cloudinit/common.go
```diff
@@ -73,7 +73,7 @@ func NewBaseCloudConfig(data BaseUserData) (CloudConfig, error) {
 
 	snapInstall := data.SnapInstallData
 	// Default to k8s version if snap install option is not set or empty.
-	if snapInstall == nil {
+	if snapInstall == nil || snapInstall.Option == "" {
 		snapInstall = &SnapInstallData{
 			Option: InstallOptionChannel,
 			Value:  fmt.Sprintf("%d.%d-classic/stable", kubernetesVersion.Major(), kubernetesVersion.Minor()),
```
bschimke95, line 79: edge case nit: What happens if the user selects a Kubernetes version where we do not (yet) have a stable release? Do we provide a good error experience for that?

bschimke95 (approved):
Generally LGTM, I think that needs to be rebased on main to include the CI fixes.

PR #94, opened 2025-01-21
Removes resource limit for controlplane manager

claudiubelu:
Setting a resource limit that is too low may lead to it crashing. Lower resource requests will make it easier for it to be scheduled.


bschimke95 (commented):
We should probably also remove the limits for the k8sd-proxy
https://github.com/canonical/cluster-api-k8s/blob/9690f71861f58633d469379e424a3ec3667d3874/pkg/ck8s/manifests/k8sd-proxy-template.yaml#L54

Issue #91, opened 2025-01-17
Support ProviderId via extra-node-kubelet-args


ader1990:
Hello,

For some infra providers, like metal3, there is the need to provide a so-called providerID via the kubelet binary parameters.

In the k8s-snap docs: https://documentation.ubuntu.com/canonical-kubernetes/latest/src/snap/reference/bootstrap-config-reference/#extra-node-kubelet-args, there is a way to achieve this by using a `extra-node-kubelet-args` bootstrap config.

But I saw that there is no configuration possible via the CK8SControlPlane to achieve the same thing.
Now I am using this way to do it, please let me know if there is an already better way or if you can implement it in the CK8SControlPlane spec maybe?

```
spec:
  spec:
    preRunCommands:
    - 'echo "extra-node-kubelet-args:" >> /capi/etc/config.yaml'
    - 'echo "  --provider-id: metal3://{{ ds.meta_data.providerid }}" >> /capi/etc/config.yaml'
```

Thank you,
Adrian.

PR #99, opened 2025-01-29
Allow passing extra service args

petrutlucian94:
The CAPI provider already allows passing extra kube-api args.

We'll do the same for the other k8s services (e.g. kubelet, kube-proxy, containerd, dqlite, etc), exposing the corresponding k8s-snap settings.

Fixes: https://github.com/canonical/cluster-api-k8s/issues/91


bschimke95 (approved):
LGTM

PR #103, opened 2025-02-11
chore: Replace hard-coded URL paths with the ones from k8s-snap-api package

rapour:
This PR suggests changes to remove the hard-coded URL path for K8sd RPC resources and replace them with the exposed variables from [K8s-snap-api](https://github.com/canonical/k8s-snap-api).

### DIFF
--- a/pkg/ck8s/workload_cluster.go
+++ b/pkg/ck8s/workload_cluster.go
@@ -201,7 +201,7 @@ func (w *Workload) GetCertificatesExpiryDate(ctx context.Context, machine *clust
 		return "", fmt.Errorf("failed to create k8sd proxy: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/ck8s/url.go
```diff
@@ -0,0 +1,11 @@
+package ck8s
+
+import (
+	"fmt"
+
+	apiv1 "github.com/canonical/k8s-snap-api/api/v1"
+)
+
+func k8sdAPIv1URL(rpcPath string) string {
+	return fmt.Sprintf("%s/%s", apiv1.K8sdAPIVersion, rpcPath)
```
bschimke95, line 10: I would be explicit here and use `fmt.Sprintf(...)` in the requests directly. 

This function hides implementation without providing a better abstraction.

bschimke95 (commented):
Generally LGTM, minor nit
Great work @rapour

pkg/ck8s/url.go
rapour: Done

bschimke95 (approved):
LGTM

Issue #104, opened 2025-02-11
Airgapped installation using LocalPath is incomplete


ader1990:
Hello,

The following workflow is required to more easily integrate `[cluster-api-k8s](https://github.com/canonical/cluster-api-k8s)` into the `https://sylva-projects.gitlab.io/` project:

  * Create an Ubuntu Image using disk-image-builder project as per Sylva requirements - to contain the k8s binaries
  * Use that Ubuntu Image with `cluster-api-k8s`
 
An issue arises with this workflow, as the disk-image-builder relies on a chroot environment, thus there is no way to pre-install a snap, in this case `k8s` snap.

A workaround would be to just download the snap files for the k8s snap: `snap download k8s`, and it is possible to do it at a known path in the image, let's say `/capi/snap-binaries` and an image to be created.

Afterwards, using the `localPath = /capi/snap-binaries/k8s.snap` in the CK8sConfig, the installation of the snap `k8s` cannot be done in an airgapped environment (airgapped environment = no access to the Internet), because the snap `k8s` is not an independent snap, as it depends on the `core20` snap.

Because of this, there is no way to actually use the `LocalPath` feature of the `cluster-api-k8s` in an airgapped environment, as it will always fail because snapd cannot download the `core20` snap.

Is this behaviour intentional? It would be great to have a way to either be able to install a snap in a chrooted environment. Or to carry over the dependent snaps with you and snap to automatically try to install those snaps from the same directory. Or to have  `cluster-api-k8s` install the `core20` snap from the root dir of the `LocalPath`.

Thank you.

claudiubelu:
This seems to work:

```
sudo snap download k8s --channel 1.32-classic/stable
sudo snap ack k8s_2187.assert
sudo snap download core20
sudo snap ack core20_2434.assert

cat /capi/etc/snap-local-path
/home/ubuntu/workdir/cluster-api-k8s/core20_2434.snap /home/ubuntu/workdir/cluster-api-k8s/k8s_2187.snap

sudo bash ./pkg/cloudinit/scripts/install.sh
k8s v1.32.1 installed
Warning: flag --classic ignored for strictly confined snap core20

core20 20240911 installed
```

Granted, a minor change needs to be applied to ``install.sh``: https://github.com/canonical/cluster-api-k8s/pull/109

PR #109, opened 2025-02-13
Allow multiple local path snaps to be installed

claudiubelu:
In airgapped environments, multiple snaps may be downloaded and then installed from local paths. The ``k8s`` snap depends on ``core20`` snap.

With this change, /capi/etc/snap-local-path may be a folder containing local snaps to install. This change preserved the old behaviour.

Fixes: https://github.com/canonical/cluster-api-k8s/issues/104

### DIFF
--- a/pkg/cloudinit/scripts/install.sh
+++ b/pkg/cloudinit/scripts/install.sh
@@ -3,7 +3,8 @@
 ## Assumptions:
 ## - /capi/etc/snap-channel contains the snap channel to be installed that matches the desired Kubernetes version, e.g. "v1.30.1" -> "1.30-classic/stable"
 ## - /capi/etc/snap-revision contains the snap revision to be installed, e.g. 123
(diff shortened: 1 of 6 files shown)


pkg/cloudinit/scripts/install.sh
```diff
@@ -13,7 +13,7 @@ elif [ -f "/capi/etc/snap-revision" ]; then
   snap install k8s --classic --revision "${snap_revision}"
 elif [ -f "/capi/etc/snap-local-path" ]; then
   snap_local_path="$(cat /capi/etc/snap-local-path)"
-  snap install k8s --classic --dangerous "${snap_local_path}"
+  snap install --classic --dangerous ${snap_local_path}
```
bschimke95, line 16: Please update the documentation/assumption above as well. Consider renaming this variable.

bschimke95 (commented):
What else is required for #104 to be resolved? Can we just add it here?

bschimke95 (approved):
LGTM

PR #120, opened 2025-03-14
test: Assert the number of cluster members on node scaling test

claudiubelu:
In the node scaling test, we are asserting the number of control plane nodes in the workload cluster after scaling the number. However, we do not check the k8sd cluster to see its view.

There have been cases in which the number of control plane nodes in the Kubernetes workload cluster does not match the number of nodes in the k8sd cluster. This will add some assertions, making sure those numbers match.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -1,0 +1,170 @@
+//go:build e2e
+// +build e2e
+
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


test/e2e/util.go
```diff
@@ -0,0 +1,170 @@
...
+	_, err = buf.ReadFrom(logStream)
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+func deletePod(clientset *kubernetes.Clientset, namespace, podName string) error {
+	return clientset.CoreV1().Pods(namespace).Delete(context.TODO(), podName, metav1.DeleteOptions{})
```
bschimke95, line 169: nit: I guess all those contexts should be tied to the test context?

bschimke95 (approved):
LGTM, we probably need to wait with the merge until #121 is merged, right?

test/e2e/util.go
claudiubelu: Done

PR #121, opened 2025-03-18
Improves how the k8sd proxy client creation process

claudiubelu:
Currently, there are a few issues with the way we create the k8sd proxy client, which can lead to the use of an invalid client, leading to errors when trying to perform k8sd-related operations.

We now make sure that the picked Pod is Ready, and that the node's k8sd API endpoint is responsive.


pkg/ck8s/util.go
```diff
@@ -0,0 +1,22 @@
+package ck8s
```
bschimke95, line 1: this should be `util`. You could also think about having `pkg/ck8s/utils/pods.go` to match the utils structure in the k8s-snap.

pkg/ck8s/util.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// isPodReady returns whether the Pod is ready or not, based on its Status.
+// based on: https://pkg.go.dev/k8s.io/kubernetes/pkg/api/v1/pod#IsPodReady
```
bschimke95, line 8: Why don't we just use this implementation then?

bschimke95 (commented):
looks good only small comments

pkg/ck8s/workload_cluster.go
```diff
@@ -157,20 +158,47 @@ func (w *Workload) GetK8sdProxyForControlPlane(ctx context.Context, options k8sd
...
 		if err != nil {
+			allErrors = append(allErrors, fmt.Errorf("could not create proxy client for node %s: %w", node.Name, err))
+			continue
+		}
+
+		// Check if there is any response from the proxy.
+		header := w.newHeaderWithCAPIAuthToken()
+		if err := w.doK8sdRequest(ctx, proxy, http.MethodGet, "", header, "", nil); err != nil {
+			allErrors = append(allErrors, fmt.Errorf("error while contacting proxy on node %s: %w", node.Name, err))
 			continue
```
HomayoonAlimohammadi, line 195: Maybe we can merge this part with the `isPodReady` function and rename the function to something like: `isProxyReady`

pkg/ck8s/util.go
```diff
@@ -0,0 +1,22 @@
...
+func isPodReady(pod *corev1.Pod) bool {
+	condition := getPodReadyCondition(&pod.Status)
+	return condition != nil && condition.Status == corev1.ConditionTrue
+}
```
HomayoonAlimohammadi, line 12: As Ben mentioned, let's use the upstream package, and also I suggest adding a function like `isProxyReady` or something like that which both checks for the readiness of the pod and a response from the proxy. We can then move it to a `pkg/ck8s/proxy` package or something like that to prevent adding more to the `util` package, as I believe is recommended by the Go team: https://go.dev/blog/package-names

HomayoonAlimohammadi (commented):
Great work, thanks a lot @claudiubelu. LGTM overall, left some comments and suggestions.

pkg/ck8s/util.go
claudiubelu: No longer applicable.

pkg/ck8s/util.go
claudiubelu: Sure, was trying to avoid adding a new dependency just for this.

pkg/ck8s/util.go
claudiubelu: No longer applicable.

pkg/ck8s/workload_cluster.go
```diff
@@ -15,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/rest"
+	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
```
bschimke95, line 19: Let's stick to the Kubernetes include conventions.
This should be `podv1`

bschimke95 (approved):
LGTM, final comment

pkg/ck8s/workload_cluster.go
claudiubelu: Done.

PR #123, opened 2025-03-24
Fix CAPI compatibility issue and bump versions

petrutlucian94:
CAPI now expects the control plane status to include the minimum K8s version of the cluster nodes.

If unset, the CAPI prechecks consider that the cluster is not ready yet and will prevent workload machines from being deployed.

https://github.com/kubernetes-sigs/cluster-api/commit/c8a5f18bc49b4f8b555129b7f5c0471f15f24856
https://github.com/kubernetes-sigs/cluster-api/blob/3cce0d973682f11ab0f0ba1c2522eba66dac2d91/internal/controllers/machineset/machineset_preflight.go#L75-L84
https://github.com/kubernetes-sigs/cluster-api/blob/3cce0d973682f11ab0f0ba1c2522eba66dac2d91/internal/controllers/machineset/machineset_preflight.go#L75-L84

This change will add the CP Version status field and set it based on the lowest K8s version identified across cluster nodes.

While at it, we'll bump the CAPI, k8s and controller-runtime versions.


controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -309,21 +309,18 @@ func (r *CK8sControlPlaneReconciler) SetupWithManager(ctx context.Context, mgr c
...
-		}
-
-		controlPlaneRef := c.Spec.ControlPlaneRef
-		if controlPlaneRef != nil && controlPlaneRef.Kind == "CK8sControlPlane" {
-			return []ctrl.Request{{NamespacedName: client.ObjectKey{Namespace: controlPlaneRef.Namespace, Name: controlPlaneRef.Name}}}
-		}
+func (r *CK8sControlPlaneReconciler) ClusterToCK8sControlPlane(_ context.Context, o client.Object) []ctrl.Request {
+	c, ok := o.(*clusterv1.Cluster)
+	if !ok {
+		panic(fmt.Sprintf("Expected a Cluster but got a %T", o))
```
berkayoz, line 315: Any particular reason why we choose to panic here instead of returning an error? I assume this is something we should never have normally?

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -355,6 +352,11 @@ func (r *CK8sControlPlaneReconciler) updateStatus(ctx context.Context, kcp *cont
 	kcp.Status.ReadyReplicas = 0
 	kcp.Status.UnavailableReplicas = replicas
 
+	lowestVersion := ownedMachines.LowestVersion()
```
berkayoz, line 355: Do we know how this `LowestVersion` is populated, is it derived from the actual node itself vs the `version` field in the cluster CRD? IIRC you can technically skip populating the version and use the snap revision for install.

petrutlucian94:
There are a few problems after the rebase, I'll fix them asap.

berkayoz (commented):
Great work, LGTM overall. Left a few questions

controlplane/controllers/ck8scontrolplane_controller.go
petrutlucian94: We're passing this function to ``EnqueueRequestsFromMapFunc`` and it doesn't allow returning an error. Indeed, we shouldn't get here normally, the tests should catch it if we pass anything other than a ``Cluster`` object.

controlplane/controllers/ck8scontrolplane_controller.go
petrutlucian94: It's actually retrieving it from the ``Machine`` spec, which I think is supposed to be an actual k8s version:

https://github.com/kubernetes-sigs/cluster-api/blob/8523bb29cf5a76a592c208c8b28fad758621fb0b/util/collections/machine_collection.go#L273-L282

berkayoz (approved):
LGTM

PR #124, opened 2025-03-25
Returns error when failing to remove machine from microcluster

claudiubelu:
If the microcluster request failed, we shouldn't proceed to remove the node's Machine, the node may still be registered as a node in the microcluster, leading it to becoming unhealthy and unstable.

### DIFF
--- a/controlplane/controllers/remediation.go
+++ b/controlplane/controllers/remediation.go
@@ -177,8 +177,13 @@ func (r *CK8sControlPlaneReconciler) reconcileUnhealthyMachines(ctx context.Cont
 		return ctrl.Result{}, fmt.Errorf("failed to create client to workload cluster: %w", err)
 	}
 
-	if err := workloadCluster.RemoveMachineFromCluster(ctx, machineToBeRemediated); err != nil {
-		log.Error(err, "failed to remove machine from microcluster")
+	if machineToBeRemediated.Status.NodeRef != nil {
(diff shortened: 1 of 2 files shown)


controlplane/controllers/remediation.go
```diff
@@ -179,6 +179,7 @@ func (r *CK8sControlPlaneReconciler) reconcileUnhealthyMachines(ctx context.Cont
 
 	if err := workloadCluster.RemoveMachineFromCluster(ctx, machineToBeRemediated); err != nil {
 		log.Error(err, "failed to remove machine from microcluster")
+		return ctrl.Result{}, fmt.Errorf("failed to remove machine from microcluster: %w", err)
```
HomayoonAlimohammadi, line 182: The only thing that concerns me is that we might get in a situation where the first time we try to remove a machine from cluster, it basically happens after the timeout of the request from CAPI. So CAPI errors, but the node removal in the k8sd side actually happens.
And from now on retries in CAPI are doomed to fail because the node is available in the k8s node list (I think we're not cleaning that up in k8sd because it should be delegated to CAPI) but trying to remove that IP from the k8sd database fails since it's not there.
Am I on the right track here? WDYT?

HomayoonAlimohammadi (commented):
God bless you for not ignoring errors. Had a question in mind tho.

bschimke95 (commented):
+1 on @HomayoonAlimohammadi comment - otherwise LGTM

bschimke95 (approved):

PR #129, opened 2025-04-09
CI: add jirasync for gh issues

louiseschmidtgen:
## Jirasync

This PR adds a jirasync automation for our github issues.

Note: Component field says k8s snap- we currently don't really filter by component. 
We could add a "CAPI k8s snap component" if we really want to but I don't see a need to.

### DIFF
--- a/.github/.jira_sync_config.yaml
+++ b/.github/.jira_sync_config.yaml
@@ -1,0 +1,34 @@
+settings:
+  # Jira project key to create the issue in
+  jira_project_key: "KU"
+
+  # Dictionary mapping GitHub issue status to Jira issue status
+  status_mapping:
+    opened: Untriaged
+    closed: done
+
+  # (Optional) Jira project components that should be attached to the created issue
+  # Component names are case-sensitive
+  components:
+    - K8s snap
+
+  # (Optional) GitHub labels. Only issues with one of those labels will be synchronized.
+  # If not specified, all issues will be synchronized
+  #  labels: []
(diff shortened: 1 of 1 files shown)


.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  add_gh_comment: false
+
+  # (Optional) (Default: true) Synchronize issue description from GitHub to Jira
+  sync_description: true
+
+  # (Optional) (Default: true) Synchronize comments from GitHub to Jira
+  sync_comments: true
+
+  # (Optional) (Default: None) Parent Epic key to link the issue to
+  epic_key: "KU-25"
```
addyess, line 29: Where does these come from?  Do we never need to change them?

addyess (approved):
LGTM.  thanks @louiseschmidtgen

.github/.jira_sync_config.yaml
HomayoonAlimohammadi: I think this is Maintenance. According to our Jira, its due date was Nov 2024. I guess for this cycle we had "Maintenance 25.04". So I think if we want to always have the parent as "Maintenance" we need to update this, or create a separate parent for these auto-created cards.

louiseschmidtgen: Yes, I would choose Maintenance 25.04 but then we need to update this workflow every cycle which isn't ideal either...

PR #130, opened 2025-04-10
fix: Disable separate feature upgrades

bschimke95:
Separate feature upgrades are currently only supported for in-place upgrades, hence disabling them for now on CAPI.


berkayoz (approved):
LGTM

Issue #126, opened 2025-04-03
*_proxy settings are not carried over to n-1 control plane nodes that need to join a multi-CP cluster


vishvikkrishnan:
[This commit](https://github.com/canonical/cluster-api-k8s/pull/69/files#diff-b5647350581d8262c9515ea238234cfb86ca54bf140674c0b75af414ab066d5d) about setting the proxy configs in the `ck8sconfig_controller.go` file skips setting the same configs on the [cloudinit](https://github.com/canonical/cluster-api-k8s/blob/main/bootstrap/controllers/ck8sconfig_controller.go#L280) for a new control plane node to join the existing cluster node. Whereas for a worker node to join, the cloud-init input does [include the proxy configs](https://github.com/canonical/cluster-api-k8s/blob/main/bootstrap/controllers/ck8sconfig_controller.go#L392-L394). This is the same case for the [init-controlplane scenario](https://github.com/canonical/cluster-api-k8s/blob/main/bootstrap/controllers/ck8sconfig_controller.go#L716-L718).

Because of this anomaly on the [join-controlplane case](https://github.com/canonical/cluster-api-k8s/blob/main/bootstrap/controllers/ck8sconfig_controller.go#L281), the *_proxy files are not written to `/capi/etc/` directory (related bug [here](https://github.com/canonical/cluster-api-k8s/issues/125)), and eventually fail the [`configure-proxy.sh` script](https://github.com/canonical/cluster-api-k8s/blob/main/pkg/cloudinit/scripts/configure-proxy.sh#L11). The *_proxy configs should also be passed on to cloudinit for a control plane node that needs to join the existing cluster node.

PR #132, opened 2025-04-11
fix: proxy settings for new CP node joining multi CP-cluster

tizianacatena:
This commit passes the *_proxy configs to cloud init for a controlplane node that needs to join the existing cluster node, to prevent *_proxy files not being written to /capi/etc/ dir.

Fixes: https://github.com/canonical/cluster-api-k8s/issues/126


petrutlucian94 (approved):

petrutlucian94:
Thanks for sending a fix!

Issue #116, opened 2025-03-04
Add support for CP rollout strategy configuration


ader1990:
Currently, there is no configuration option for the rollout strategy regarding CK8SControlPlane.

Would be great to have a way to configure it in the future, similar to how kubeadm or rke2 cluster api providers implement it - https://github.com/rancher/cluster-api-provider-rke2/blob/main/controlplane/api/v1beta1/rke2controlplane_types.go#L99.

Example spec:

```
spec:
  strategy:
      rollingUpdate:
        maxSurge: 0
      type: RollingUpdate
```

Thanks.

PR #135, opened 2025-04-15
Add support for CP rollout strategy configuration

ader1990:
Fixes: https://github.com/canonical/cluster-api-k8s/issues/116

The CK8sControlPlane will have the extra configuration option to set the maxSurge spec.strategy.rollingUpdate.maxSurge to 0.

```
spec:
  strategy:
      rollingUpdate:
        maxSurge: 1 | 0
```

By default the maxSurge is set to value 1, meaning that when a cluster rollingUpdate is performed, the ControlPlanes are rolled out in the N, N+1, N fashion (where N is the number of total Control Planes). This workflow always required one extra spare Machine to be available during the rolling update.

If the maxSurge is set to value 0, meaning that when a cluster rollingUpdate is performed, the ControlPlanes are rolled out in the N, N-1, N fashion (where N is the number of total Control Planes). This workflow does not require one Machine to be available during the rolling update.


controlplane/api/v1beta2/ck8scontrolplane_webhook.go
```diff
@@ -76,4 +77,29 @@ func defaultCK8sControlPlaneSpec(s *CK8sControlPlaneSpec, namespace string) {
...
+	// Enforce RollingUpdate strategy and default MaxSurge if not set.
+	if rolloutStrategy != nil {
+		if len(rolloutStrategy.Type) == 0 {
+			rolloutStrategy.Type = RollingUpdateStrategyType
+		}
+		if rolloutStrategy.Type == RollingUpdateStrategyType {
+			if rolloutStrategy.RollingUpdate == nil {
+				rolloutStrategy.RollingUpdate = &RollingUpdate{}
+			}
+			rolloutStrategy.RollingUpdate.MaxSurge = intstr.ValueOrDefault(rolloutStrategy.RollingUpdate.MaxSurge, ios1)
```
claudiubelu, line 100: question: is the given `rolloutStrategy.Type` verified anywhere? What if I have typo? Same with `MaxSurge` - we say we support the values of 0 and 1, but what happens if we're given `MaxSurge` -2, or 5?

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -738,10 +738,25 @@ func (r *CK8sControlPlaneReconciler) upgradeControlPlane(
...
 
-	if controlPlane.Machines.Len() <= int(*kcp.Spec.Replicas) {
-		// scaleUp ensures that we don't continue scaling up while waiting for Machines to have NodeRefs
-		return r.scaleUpControlPlane(ctx, cluster, kcp, controlPlane)
+	if kcp.Spec.RolloutStrategy == nil {
+		logger.Info("RolloutStrategy is empty, unable to continue")
+		return ctrl.Result{}, nil
+	}
+
+	switch kcp.Spec.RolloutStrategy.Type {
```
ader1990, line 748: I am getting here a `"panicGoValue": "\"invalid memory address or nil pointer dereference\"",` -> which means I am not doing the correct defaulting. Is there something I am missing?

ader1990: This has been resolved by adding the default value via annotations.

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -738,10 +738,20 @@ func (r *CK8sControlPlaneReconciler) upgradeControlPlane(
...
+		// We can ignore MaxUnavailable because we are enforcing health checks before we get here.
+		maxNodes := *kcp.Spec.Replicas + int32(kcp.Spec.RolloutStrategy.RollingUpdate.MaxSurge.IntValue())
+		if int32(controlPlane.Machines.Len()) < maxNodes {
+			// scaleUp ensures that we don't continue scaling up while waiting for Machines to have NodeRefs
+			return r.scaleUpControlPlane(ctx, cluster, kcp, controlPlane)
+		}
+		return r.scaleDownControlPlane(ctx, cluster, kcp, controlPlane, machinesRequireUpgrade)
+	default:
+		logger.Info("RolloutStrategy type is not set to RollingUpdateStrategyType, unable to determine the strategy for rolling out machines")
+		return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
ader1990: This is the upstream implementation from kubeadm.
What misses here is the proper validation to all the specs, which should be done for all the configs, and not only for this one that I am adding now.
See: https://github.com/kubernetes-sigs/cluster-api/blob/dd105a3d9059f54dd09f6c278367157540a883a2/controlplane/kubeadm/internal/webhooks/kubeadm_control_plane.go#L393

I would either leave this with no validation or add a new PR, separate from this, that adds proper validation to all the validatable configurations.

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -738,10 +738,20 @@ func (r *CK8sControlPlaneReconciler) upgradeControlPlane(
...
 
-	if controlPlane.Machines.Len() <= int(*kcp.Spec.Replicas) {
-		// scaleUp ensures that we don't continue scaling up while waiting for Machines to have NodeRefs
-		return r.scaleUpControlPlane(ctx, cluster, kcp, controlPlane)
+	switch kcp.Spec.RolloutStrategy.Type {
+	case controlplanev1.RollingUpdateStrategyType:
+		// RolloutStrategy is currently defaulted and validated to be RollingUpdate
+		// We can ignore MaxUnavailable because we are enforcing health checks before we get here.
+		maxNodes := *kcp.Spec.Replicas + int32(kcp.Spec.RolloutStrategy.RollingUpdate.MaxSurge.IntValue())
+		if int32(controlPlane.Machines.Len()) < maxNodes {
```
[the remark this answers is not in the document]
ader1990: Yes, the maxSurge is set in a backwards compatible fashion with value 1. Which means, that when a rolling upgrade happens, the workflow starts with an upscale and then a downscale. Setting maxSurge = 0, the workflow starts with a downscale of one CP node, then upscale of one CP node.

controlplane/api/v1beta2/ck8scontrolplane_webhook.go
```diff
@@ -76,4 +77,29 @@ func defaultCK8sControlPlaneSpec(s *CK8sControlPlaneSpec, namespace string) {
...
+func defaultRolloutStrategy(rolloutStrategy *RolloutStrategy) *RolloutStrategy {
+	ios1 := intstr.FromInt(1)
+
+	if rolloutStrategy == nil {
+		rolloutStrategy = &RolloutStrategy{}
+	}
+
+	// Enforce RollingUpdate strategy and default MaxSurge if not set.
+	if rolloutStrategy != nil {
+		if len(rolloutStrategy.Type) == 0 {
```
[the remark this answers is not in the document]
ader1990: Because of how types are in yaml, and can be empty or undefined, I think the len version was prefered upstream to be more safe.

controlplane/api/v1beta2/ck8scontrolplane_webhook.go
ader1990: ditto from below.

controlplane/api/v1beta2/ck8scontrolplane_webhook.go
```diff
@@ -76,4 +77,29 @@ func defaultCK8sControlPlaneSpec(s *CK8sControlPlaneSpec, namespace string) {
...
+
+func defaultRolloutStrategy(rolloutStrategy *RolloutStrategy) *RolloutStrategy {
+	ios1 := intstr.FromInt(1)
+
+	if rolloutStrategy == nil {
+		rolloutStrategy = &RolloutStrategy{}
+	}
+
+	// Enforce RollingUpdate strategy and default MaxSurge if not set.
+	if rolloutStrategy != nil {
```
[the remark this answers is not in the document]
ader1990: This code was added from upstream, I would leave it as is for better comparison / diff when needed.

controlplane/api/v1beta2/ck8scontrolplane_types.go
```diff
@@ -22,11 +22,22 @@ import (
...
 )
 
+// RolloutStrategyType defines the rollout strategies for a CK8sControlPlane.
+// +kubebuilder:validation:Enum=RollingUpdate
+type RolloutStrategyType string
+
+const (
+	// RollingUpdateStrategyType replaces the old control planes by new one using rolling update
+	// i.e. gradually scale up or down the old control planes and scale up or down the new one.
+	RollingUpdateStrategyType RolloutStrategyType = "RollingUpdate"
```
ktsakalozos, line 38: If there is only one RolloutStrategy there is no point in having this option.

ktsakalozos: I would skip the type:

```
spec:
  strategy:
      rollingUpdate:
        maxSurge: 0
```

controlplane/api/v1beta2/ck8scontrolplane_types.go
```diff
@@ -109,6 +116,26 @@ type CK8sControlPlaneMachineTemplate struct {
...
+// RollingUpdate is used to control the desired behavior of rolling update.
+type RollingUpdate struct {
+	// maxSurge is the maximum number of control planes that can be scheduled above or under the
+	// desired number of control planes.
+	// Value can be an absolute number 1 or 0.
+	// Defaults to 1.
+	// Example: when this is set to 1, the control plane can be scaled
+	// up immediately when the rolling update starts.
+	// +optional
+	MaxSurge *intstr.IntOrString `json:"maxSurge,omitempty"`
```
berkayoz, line 136: I honestly do not like that upstream also uses `intstr.IntOrString`, an `int` would enable us to have proper CRD validation. Do we know if we can use `kubebuilder:validation:XValidation` in anyway with `IntOrString`?

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -738,8 +738,15 @@ func (r *CK8sControlPlaneReconciler) upgradeControlPlane(
 		return reconcile.Result{}, fmt.Errorf("failed to upgrade kubelet config map")
 	}
 	**/
+	logger := r.Log.WithValues("namespace", kcp.Namespace, "CK8sControlPlane", kcp.Name, "cluster", cluster.Name)
+
+	if kcp.Spec.RolloutStrategy == nil {
+		logger.Info("RolloutStrategy is empty, unable to continue")
+		return ctrl.Result{}, nil
+	}
```
berkayoz, line 746: Shouldn't this return an error? Do we want to abort the rollout process totally without retrying?

berkayoz (commented):
Hey @ader1990, great work! Did an initial pass and left some comments to clarify some things.

controlplane/api/v1beta2/ck8scontrolplane_types.go
ader1990: Hello, I do not know as I am not so familiar with those annotations.

controlplane/config/manager/kustomization.yaml
```diff
@@ -5,3 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: ghcr.io/canonical/cluster-api-k8s/controlplane-controller
+  newTag: v0.4.1
```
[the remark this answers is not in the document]
ader1990: this is a leftover from the autobuild process I added to the code. removed.

controlplane/controllers/ck8scontrolplane_controller.go
ader1990: This was an extra check I had to do `just in case`, before setting a default value. I think it might be safe to remove the check.

ader1990: I would still leave the code as is, and not return an error and abort the process - because if the configuration is broken, then it should be aborted.

berkayoz (approved):
LGTM @ader1990, thanks. Can we rebase on `main` to fix CI failures?

ader1990:
> LGTM @ader1990, thanks. Can we rebase on `main` to fix CI failures?

Hello, the PR was rebased, if you can please trigger the CI (as I do not have the rights)?

PR #140, opened 2025-04-22
test: Adds Upgrade test scenario with MaxSurge=0

claudiubelu:
Currently, we only test the rolling upgrade with the default ``MaxSurge=1`` (for a 3-node control plane cluster, a 4th node is created and joined before removing one of the nodes).

This adds a test for the rolling upgrade with ``MaxSurge=0`` scenario (for a 3-node control plane cluster, it first removes a node, then
creates and joins a new one). The test ensures that the numbers of Machines present in CAPI is only 3 for this scenario.


claudiubelu:
Closing in favor of: https://github.com/canonical/cluster-api-k8s/pull/140 . The PR contains the test introduced in this PR.

PR #152, opened 2025-06-01
fix(cluster-template/CK8sControlPlane): use spec.channel

hansbogert:
This was needed for the demo to work.


rapour (approved):
Thanks a lot for fixing the template, LGTM

@hansbogert Could you please rebase from main to fix the CI?

PR #155, opened 2025-06-25
fix: use TiCS action and new URL

rapour:
TIOBE recently changed the URL for the projects that use Golang. This pull request updates the URL and switches to a different workflow, as recommended by other projects.


rapour:
Closing in favor of #157

PR #157, opened 2025-07-01
fix: separate tics and use new action

rapour:
TIOBE recently changed the URL for the projects that use Golang. This pull request updates the URL and switches to a different workflow, as recommended by other projects.


ktsakalozos (approved):
LGTM +1

ktsakalozos:
The CI error is not related to this PR and will be addressed in another PR.

PR #160, opened 2025-07-11
feat: add managed etcd datastore option

berkayoz:
Should be rebased and merged after #171 and #172

### DIFF
--- a/bootstrap/api/v1beta2/ck8sconfig_types.go
+++ b/bootstrap/api/v1beta2/ck8sconfig_types.go
@@ -136,7 +136,7 @@ type CK8sConfigSpec struct {
 // IsEtcdManaged returns true if the control plane is using k8s-dqlite.
 func (c *CK8sConfigSpec) IsEtcdManaged() bool {
 	switch c.ControlPlaneConfig.DatastoreType {
-	case "", "k8s-dqlite":
+	case "", "etcd", "k8s-dqlite":
 		return true
 	}
 	return false
@@ -167,6 +167,14 @@ type CK8sControlPlaneConfig struct {
 	// K8sDqlitePort is the port to use for k8s-dqlite. If unset, 2379 (etcd) will be used.
 	// +optional
 	K8sDqlitePort int `json:"k8sDqlitePort,omitempty"`
+
+	// EtcdPort is the port to use for etcd. If unset, 2379 will be used.
+	// +optional
+	EtcdPort int `json:"etcdPort,omitempty"`
+
+	// EtcdPeerPort is the port to use for etcd peer communication. If unset, 2381 will be used.
+	// +optional
+	EtcdPeerPort int `json:"etcdPeerPort,omitempty"`
 
 	// MicroclusterAddress is the address (or CIDR) to use for microcluster. If unset, the default node interface is chosen.
 	MicroclusterAddress string `json:"microclusterAddress,omitempty"`
--- a/pkg/ck8s/config_init.go
+++ b/pkg/ck8s/config_init.go
@@ -70,7 +70,23 @@ func GenerateInitControlPlaneConfig(cfg InitControlPlaneConfig) (apiv1.Bootstrap
 	}
 
 	switch cfg.DatastoreType {
-	case "", "k8s-dqlite":
+	case "", "etcd":
+		out.DatastoreType = ptr.To("etcd")
+
+		etcdPort := cfg.ControlPlaneConfig.EtcdPort
+		if etcdPort == 0 {
+			etcdPort = 2379
+		}
+		out.EtcdPort = ptr.To(etcdPort)
+
+		etcdPeerPort := cfg.ControlPlaneConfig.EtcdPeerPort
+		if etcdPeerPort == 0 {
+			// TODO(berkayoz): This should be 2080 however it clashes with our workaround
+			// for exposing microcluster through 2080 via k8sd-proxy
+			etcdPeerPort = 2381
+		}
+		out.EtcdPeerPort = ptr.To(etcdPeerPort)
+	case "k8s-dqlite":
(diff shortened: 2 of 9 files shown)


controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -561,12 +561,22 @@ func (r *CK8sControlPlaneReconciler) reconcile(ctx context.Context, cluster *clu
...
 	}
 
-	obj, err := external.Get(ctx, r.Client, &ref)
+	logger := r.Log.WithValues("namespace", ref.Namespace, "CK8sControlPlane", ref.Name, "cluster", cluster.Name)
+	logger.Info("Reconciling external template reference", "ref", ref)
+
+	// Ensure the ref namespace is populated for objects not yet defaulted by webhook
+	// https://github.com/kubernetes-sigs/cluster-api/pull/11361
+	if ref.Namespace == "" {
+		ref = ref.DeepCopy()
```
HomayoonAlimohammadi, line 575: curio: why?

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -561,12 +561,22 @@ func (r *CK8sControlPlaneReconciler) reconcile(ctx context.Context, cluster *clu
...
+func (r *CK8sControlPlaneReconciler) reconcileExternalReference(ctx context.Context, cluster *clusterv1.Cluster, ref *corev1.ObjectReference) error {
 	if !strings.HasSuffix(ref.Kind, clusterv1.TemplateSuffix) {
 		return nil
 	}
 
-	obj, err := external.Get(ctx, r.Client, &ref)
+	logger := r.Log.WithValues("namespace", ref.Namespace, "CK8sControlPlane", ref.Name, "cluster", cluster.Name)
+	logger.Info("Reconciling external template reference", "ref", ref)
+
+	// Ensure the ref namespace is populated for objects not yet defaulted by webhook
```
HomayoonAlimohammadi, line 572: I'm asking this because I don't know how this exactly works. How is the webhook going to default the object? If it only populates fields without a value with some default value, it's not gonna default "namespace" anymore since we're populating it ourselves, right?
I'm just worried that this might cause a behavior that we're not interested in. Do we rely on the default namespace that gets assigned by the webhook?

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -561,12 +561,22 @@ func (r *CK8sControlPlaneReconciler) reconcile(ctx context.Context, cluster *clu
 	return reconcile.Result{}, nil
 }
 
-func (r *CK8sControlPlaneReconciler) reconcileExternalReference(ctx context.Context, cluster *clusterv1.Cluster, ref corev1.ObjectReference) error {
+func (r *CK8sControlPlaneReconciler) reconcileExternalReference(ctx context.Context, cluster *clusterv1.Cluster, ref *corev1.ObjectReference) error {
```
HomayoonAlimohammadi, line 564: curio: why?

pkg/ck8s/config_init.go
```diff
@@ -70,7 +70,23 @@ func GenerateInitControlPlaneConfig(cfg InitControlPlaneConfig) (apiv1.Bootstrap
 	}
 
 	switch cfg.DatastoreType {
-	case "", "k8s-dqlite":
+	case "", "etcd":
+		out.DatastoreType = ptr.To("etcd")
```
HomayoonAlimohammadi, line 74: nit: Can we extract the "etcd" string to a const?

bootstrap/api/v1beta2/ck8sconfig_types.go
```diff
@@ -168,6 +168,14 @@ type CK8sControlPlaneConfig struct {
 	// +optional
 	K8sDqlitePort int `json:"k8sDqlitePort,omitempty"`
 
+	// EtcdPort is the port to use for etcd. If unset, 2379 will be used.
+	// +optional
+	EtcdPort int `json:"etcdPort,omitempty"`
+
+	// EtcdPeerPort is the port to use for etcd peer communication. If unset, 2380 will be used.
```
HomayoonAlimohammadi, line 175: I think we should change this to 2381 due to the MC conflict workaround below.

pkg/ck8s/config_init.go
HomayoonAlimohammadi: I noticed that we're not doing this for k8s-dqlite as well, so please either do this for both or just ignore this comment. It's a nit so I'm OK with both.

templates/docker/Dockerfile
```diff
@@ -20,12 +20,12 @@
 ## See https://kind.sigs.k8s.io/docs/design/node-image/
 ##
 ## !!!IMPORTANT!!! Keep this up to date with the minor version of k8s-snap.
```
HomayoonAlimohammadi, line 22: The fact that this is IMPORTANT yet not automated is a bummer. Let me create a card for this.

HomayoonAlimohammadi: KU-3966

test/e2e/intermediate_ca_test.go
HomayoonAlimohammadi, line 1: I wonder why/how indentations in this file were increased to 2 levels without Go fixing it itself.

HomayoonAlimohammadi (approved):
Great work, thanks a lot @berkayoz! LGTM overall, left some questions as well.

bschimke95 (commented):
There are multiple fixes in this PR on top of the etcd datastore changes which makes it a bit though to review. 
Please move them into a separate PR.

controlplane/controllers/ck8scontrolplane_controller.go
berkayoz: Moved to #172

PR #166, opened 2025-07-21
fix: run tmate session in detached mode before running the tests

rapour:


HomayoonAlimohammadi (approved):
LGTM, thanks! We need to move the tmate step before the tests so that we have a large enough window to debug.

PR #174, opened 2025-08-07
ci: fix prereleases ci

louiseschmidtgen:
## Description

This PR fixes the pre-releases CI, the workflow was missing write permissions to create and push the tags.


addyess (approved):
Thanks!

Issue #183, opened 2025-10-06
Rolling upgrades (rolling updates in CAPI term) lead to forever stuck pods in `Terminating` state and nodes stuck in `NotReady,SchedulingDisabled`


ader1990:
Hello,

In the context of deploying Canonical Kubernetes with Sylva (https://sylva-projects.gitlab.io/), during rolling upgrades where MaxSurge=0, and etcd used as a data-store, the rolling upgrade gets stuck during the initial phase of moving the workloads from the to-be removed nodes to the remaining ones.

Context:

1. Deploy a 4 node (3 CP and 1 MD) Canonical Kubernetes cluster, using the https://github.com/canonical/cluster-api-k8s/commit/ca2509f4115d47d859b76f44d73eb0e8c6f3dab1 image and metal3 as infra provider. Everything works as expected.
2. Trigger a rolling upgrade via CAPI, using MaxSurge=0. 1 CP and 1 MD are set into `SchedulingDisabled` and the workloads from those two nodes should move to the remaining 2 CPs. Sometimes, the 1 CP and 1 MD are stuck with the status `NotReady`, without the `SchedulingDisabled` taint.

Expected result: the workloads should move to the remaining 2 CPs and the Kubernetes cluster should end up consisting of only 2 CPs (intermediary state, as afterwards another CP and another MD is provisioned.

Actual result: the workloads do not move, as the pods are forever stuck in `Terminating` state. The 1 CP and the 1 MD remain forever stuck in `NotReady,SchedulingDisabled` state.

Reproduceability: always.

OS used: Ubuntu Server 24.04 with latest updates.
k8s-snap Canonical Kubernetes used: 1.32.8, 1.32.9 with etcd. The variant with dqlite deployed does not show this issue (the dqlite variant fails to perform the rolling upgrades with other behaviour, unrelated to this one).

Investigation done: it was observed that on the to-be deleted CP the underlying k8s-snap etcd process does get killed very soon, way before the workloads have time to move, and the k8s status shows the node as not bootstrapped. It seems that the actual cleanup it is done in a very eager manner, and the Kubernetes cluster does not have the time to move the workloads or to respond to node removals.

<img width="2836" height="1608" alt="Image" src="https://github.com/user-attachments/assets/d1d47023-5a1c-481a-8082-5bb8e3059a78" />

<img width="1679" height="203" alt="Image" src="https://github.com/user-attachments/assets/3759a77b-aa64-49d0-ab82-5e38cd084248" />

[...]

PR #184, opened 2025-10-20
fix: Unblock CI by running the tests using the LXD infra provider

HomayoonAlimohammadi:
### Overview

This PR adds support for running the tests using the LXD infra provider

### Notes

This [this commit](1d13bd157c11cacf38d56612f6833af22449a747) we're removing the `SkipServiceStop` annotation and providing `nodeXTimeout` values which will get passed to the created `Machine`s. 
The skip annotation was originally introduced to prevent the snap services from being stopped upon removal, because the core CAPI provider performs a cordon and drain operation on the node-to-be-removed, so we needed some services to be up an running for that to happen.
But at the same time, keeping the API server alive, caused a set of new issues to surface (e.g. requests to the control plane LB getting redirected to the dead API server, or the dead API server IP not getting removed from the LB config like that of HAProxy, etc.) So we decided to pass sensible `nodeXTimeout` values to machines to instruct the core CAPI provider to timeout those operations. This is the approach that upstream is also taking. 

See [this](https://github.com/canonical/cluster-api-k8s/pull/184#discussion_r2480354636) and [this](https://github.com/canonical/cluster-api-k8s/pull/184#discussion_r2480474660) as well.

### Addresses

- https://github.com/canonical/cluster-api-k8s/issues/183
- https://github.com/canonical/k8s-snap/issues/1966

### Future work

Creating the bootstrap cluster can be supported natively (the way it is for CAPD), when using CAPN.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,9 +104,6 @@ E2E_CONF_FILE ?= $(TEST_DIR)/e2e/config/ck8s-$(E2E_INFRA).yaml
 SKIP_RESOURCE_CLEANUP ?= false
 SKIP_BOOTSTRAP_CLUSTER_INITIALIZATION ?= false
 USE_EXISTING_CLUSTER ?= false
-# TODO(Hue): Support initializing bootstrap provider as a test step
-# EXISTING_CLUSTER_KUBECONFIG_PATH ?= $(HOME)/.kube/config
-# PROVIDER_IMAGES_TAR_PATH ?= $(shell pwd)/provider-images.tar
 GINKGO_NOCOLOR ?= false
 
 # to set multiple ginkgo skip flags, if any
@@ -294,8 +291,6 @@ test-e2e: $(GINKGO) $(KUSTOMIZE) ## Run the end-to-end tests
 	    -e2e.skip-resource-cleanup=$(SKIP_RESOURCE_CLEANUP) \
 		-e2e.use-existing-cluster=$(USE_EXISTING_CLUSTER) \
 		-e2e.skip-bootstrap-cluster-initialization=$(SKIP_BOOTSTRAP_CLUSTER_INITIALIZATION)
-# 		-e2e.existing-cluster-kubeconfig-path=$(EXISTING_CLUSTER_KUBECONFIG_PATH) \
-# 		-e2e.provider-images-tar-path=$(PROVIDER_IMAGES_TAR_PATH)
 
 # Build manager binary
 manager-controlplane: generate-controlplane


.github/workflows/e2e.yaml
```diff
@@ -94,42 +80,37 @@ jobs:
...
-      - name: Create docker network
-        run: |
-          sudo docker network create kind --driver=bridge -o com.docker.network.bridge.enable_ip_masquerade=true
-      - name: Increase inotify watches
+      - name: Setup bootstrap cluster
         run: |
-          # Prevents https://cluster-api.sigs.k8s.io/user/troubleshooting#cluster-api-with-docker----too-many-open-files
-          sudo sysctl fs.inotify.max_user_watches=1048576
-          sudo sysctl fs.inotify.max_user_instances=8192
+          sudo ./hack/setup-bootstrap-cluster.sh bootstrap-cluster 1.32 ./provider-images.tar
```
berkayoz: I assume the `provider-images.tar` is passed so it can be side-loaded into management cluster right?

pkg/token/token.go
```diff
@@ -64,15 +64,20 @@ func randomB64(size int) (string, error) {
...
 
-// name returns the name of the token secret, computed by convention using the name of the cluster.
-func name(clusterName string) string {
+// secretName returns the name of the token secret, computed by convention using the name of the cluster.
+func secretName(clusterName string) string {
 	return fmt.Sprintf("%s-token", clusterName)
 }
 
+// machineNodeTokenEntry returns the key used to store the node token for a specific machine in the secret data.
+func machineNodeTokenEntry(machineName string) string {
```
berkayoz: nit: What do you think about `machineNodeTokenKey` ?

test/e2e/data/infrastructure-aws/cluster-template.yaml
```diff
@@ -51,6 +51,9 @@ spec:
       apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
       kind: AWSMachineTemplate
       name: ${CLUSTER_NAME}-control-plane
+    nodeDrainTimeout: 30s
+    nodeVolumeDetachTimeout: 30s
+    nodeDeletionTimeout: 30s
```
berkayoz: Do we set any defaults if unspecified for these?

test/e2e/intermediate_ca_test.go
```diff
@@ -18,181 +18,184 @@ limitations under the License.
...
-    "time"
-
-    . "github.com/onsi/ginkgo/v2"
-    . "github.com/onsi/gomega"
-    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-    corev1 "k8s.io/api/core/v1"
-    "k8s.io/utils/pointer"
-    "sigs.k8s.io/cluster-api/test/framework/clusterctl"
-    "sigs.k8s.io/cluster-api/util"
+	"context"
```
berkayoz: This file is a bit interesting do we know why the diff is messed up, it is hard to make out the actual changes.

test/e2e/md_remediation_test.go
```diff
@@ -60,6 +60,9 @@ var _ = Describe("When testing MachineDeployment remediation", func() {
 		// Setup a Namespace where to host objects for this spec and create a watcher for the namespace events.
 		namespace, cancelWatches = setupSpecNamespace(ctx, specName, bootstrapClusterProxy, artifactFolder)
 
+		// Create LXC secret for LXD provider if needed
+		createLXCSecret(ctx, bootstrapClusterProxy, e2eConfig, namespace.Name)
```
berkayoz: Well ultimately I don't like having this repeat for every single test although I don't see a quicker or easier alternative, maybe something with webhooks can be created to inject this in some other way after namespace creation. No action needed for now.

berkayoz (commented):
LGTM mostly, left a couple of small comments. Amazing work, cheers!

test/e2e/data/infrastructure-aws/cluster-template.yaml
HomayoonAlimohammadi: No we don't. Do you think we should?

test/e2e/intermediate_ca_test.go
HomayoonAlimohammadi: Wow, let me see what happened.

pkg/token/token.go
HomayoonAlimohammadi: Done

test/e2e/node_scale_test.go
```diff
@@ -86,22 +89,14 @@ var _ = Describe("Workload cluster scaling", func() {
...
-					WorkerMachineCount:       pointer.Int64Ptr(1),
+					ControlPlaneMachineCount: ptr.To(int64(1)),
+					WorkerMachineCount:       ptr.To(int64(1)),
 				},
 				WaitForClusterIntervals:      e2eConfig.GetIntervals(specName, "wait-cluster"),
 				WaitForControlPlaneIntervals: e2eConfig.GetIntervals(specName, "wait-control-plane"),
 				WaitForMachineDeployments:    e2eConfig.GetIntervals(specName, "wait-worker-nodes"),
 			}, result)
 
-			// Check that the number of control plane cluster members matches what we expect.
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: Added a link to this comment in the PR description.

test/e2e/data/infrastructure-aws/cluster-template.yaml
HomayoonAlimohammadi: I only see a default of 10s for `nodeDeletionTimeout`: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/core/v1beta2/machine_types.go#L459-L477

HomayoonAlimohammadi: But this is a default that's being set on the MachinPool webhook. So I don't think we need to take any action here. WDYT?

PR #185, opened 2025-10-21
feat: Cordon and drain the node before removing from the cluster

HomayoonAlimohammadi:
### Overview

This PR will cordon and drain the node that is to be deleted (during a scaledown, etc.)
If a failure occurs, the node is uncordoned to prevent having unschedulable nodes that are not going to be deleted.
The draining process is retried until there's no other pod on the node that can be evicted/deleted.
Note that we can not retry pod eviction through retriggering the reconciliation, since each time a new machine can be selected for deletion.

### DIFF
--- a/pkg/ck8s/drainer.go
+++ b/pkg/ck8s/drainer.go
@@ -1,0 +1,332 @@
+package ck8s
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


HomayoonAlimohammadi:
Closing this in favor of https://github.com/canonical/cluster-api-k8s/pull/184. We also prefer to perform a drain operation on the snap side during removal in order to be decoupled from the upper layer. This can be implemented in a future work on the snap side.

PR #216, opened 2026-04-07
ci: pin GitHub Actions to commit SHAs

louiseschmidtgen:
## Summary

Pin all GitHub Actions to full commit SHAs for supply-chain security hardening, replacing mutable tag references.

## Why

Mutable tags (e.g. `@v4`) can be moved by upstream maintainers or by an attacker who compromises an action repo. Pinning to a SHA ensures workflows always run exactly the reviewed code.

This follows GitHub's own security hardening guidance and improves our OpenSSF Scorecard rating.\n\n## Changes\n\n- Pinned all third-party and GitHub-maintained actions in workflow files to full 40-character commit SHAs\n- Preserved original tag versions in inline comments for readability\n- Left already-pinned actions unchanged\n\n## Testing\n\n- Verified all `uses:` references in `.github/workflows/` now use immutable SHAs

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -19,16 +19,16 @@ jobs:
     runs-on: [self-hosted, linux, X64, jammy, xlarge]
     steps:
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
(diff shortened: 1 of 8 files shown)


bschimke95 (approved):
All GitHub Action SHAs verified against their claimed version tags. Changes are correct and consistent.

Issue #169, opened 2025-07-28
Canonical K8s MAAS cluster creation with CAPI gets stuck when default CNI is set to false


vishu2498:
### Issue:

When deploying a Canonical K8s MAAS cluster using CAPI, if the default CNI option is set to false, the managment cluster is forever waiting for the bootstrap cluster to get created and arrive in running state since the CNI needs to be installed manually on the MAAS instance.

### Steps to reproduce:

An option is present to set default CNI to false as shown here:
```
spec:
  initConfig:
    enableDefaultNetwork: false
```

Reference: https://documentation.ubuntu.com/canonical-kubernetes/latest/capi/reference/configs/#initconfig

Create a cluster using this configuration in `CK8ControlPlane` object and the issue will be reproducible.

### Expected Behaviour:

The bootstrap cluster inside MAAS instance should be up and running and setting default CNI to false should not impact the boostrap cluster. Later when the bootstrap cluster is up (possibly using default CNI), user/CAPI can install their own CNI on the created cluster.

### Actual Behaviour:

Inside the config used by script, it disables the CNI installation (cilium CNI installation). This methods works correct when deploying with `k8s-snap` [package](https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/networking/alternative-cni/), but it fails via the CAPI method.

The bootstrap cluster is created without CNI and is stuck with only these pods:
```
NAMESPACE     NAME                              READY   STATUS    RESTARTS   AGE
kube-system   coredns-fc9c778db-bhqwc           0/1     Pending   0          11s
kube-system   metrics-server-8694c96fb7-6b79j   0/1     Pending   0          11s
```

At this stage, `k8s status` doesn't show ready status and stays the same.

From the [CAPMAAS](https://github.com/spectrocloud/cluster-api-provider-maas/blob/6ec0c47f173da23fe1b4935a7823013a9eb0ff15/pkg/maas/scope/cluster.go#L224) side, it only proceeds if it is able to create a client to target cluster (using kubeconfig secret) and is able to list the nodes (only then it confirms that API server is online).

It waits with logs like this:
```
I0717 05:05:24.427993       1 maasmachine_controller.go:353] "API Server is not online; requeue" logger="controllers.MaasMachine" maasmachine="vishu-maas-1-cp-61386-8vvln" cluster="vishu-maas-1" state="Deploying" m-id="aqhn73"
```

[...]

PR #224, opened 2026-05-20
Update Canonical K8s CAPI cluster to get ready when enableDefaultNetwork is set as false

vishu2498:
This PR lets CK8s mark the `CK8sControlPlane` object as `Ready` when default CNI is disabled in spec:
```
spec:
  initConfig:
    enableDefaultNetwork: false
```

It also lets `k8sd-proxy` daemonset to get scheduled on the `NotReady` node so that once the user installs CNI on the CAPI Canonical K8s cluster, the `k8sd-proxy` daemonset will also be up and running.

An elongated discussion is present here and this PR marks the resolution of the issue: https://github.com/canonical/cluster-api-k8s/issues/169

### DIFF
--- a/controlplane/controllers/ck8scontrolplane_controller.go
+++ b/controlplane/controllers/ck8scontrolplane_controller.go
@@ -395,12 +395,24 @@ func (r *CK8sControlPlaneReconciler) updateStatus(ctx context.Context, kcp *cont
 	kcp.Status.ReadyReplicas = status.ReadyNodes
 	kcp.Status.UnavailableReplicas = replicas - status.ReadyNodes
 
-	// NOTE(neoaggelos): We consider the control plane to be initialized iff the k8sd-config exists
-	if status.HasK8sdConfigMap {
+	enableDefaultNetwork := kcp.Spec.CK8sConfigSpec.InitConfig.GetEnableDefaultNetwork()
+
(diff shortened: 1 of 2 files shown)


controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -395,12 +395,24 @@ func (r *CK8sControlPlaneReconciler) updateStatus(ctx context.Context, kcp *cont
 	kcp.Status.ReadyReplicas = status.ReadyNodes
 	kcp.Status.UnavailableReplicas = replicas - status.ReadyNodes
 
-	// NOTE(neoaggelos): We consider the control plane to be initialized iff the k8sd-config exists
-	if status.HasK8sdConfigMap {
+	enableDefaultNetwork := kcp.Spec.CK8sConfigSpec.InitConfig.GetEnableDefaultNetwork()
```
ader1990, line 398: Looks good to me.

Needs to be tested in an end to end scenario though to make sure it works as expected.

Also, would be nice to add that this change, if I am correct to assume, is only temporary after the deployment, as CAPI controller will revert back the node states in NotReady after some time, if a CNI is not deployed?

HomayoonAlimohammadi (approved):
LGTM, thanks a lot @vishu2498!

PR #226, opened 2026-06-10
WIP: add rolling upgrade fixes

ader1990:
Work in progress.

Add the following new workflows:

  * cherry-picked @HomayoonAlimohammadi's work from https://github.com/canonical/cluster-api-k8s/pull/185 that adds node drain/cordon before removal
  * fixed drain/cordon unit tests
  * use a soft node removal request to k8sd and fallback on a forced node removal if the soft removal fails
  * add more complex up and down end to end tests


pkg/ck8s/management_cluster.go
```diff
@@ -98,12 +98,24 @@ func (m *Management) GetWorkloadCluster(ctx context.Context, clusterKey client.O
 		return nil, fmt.Errorf("auth token not yet generated")
 	}
 
+	drainer := NewDrainer(c, time.Now, DrainOptions{
+		Force:                 true,
+		AllowDeletion:         true,
+		IgnoreDaemonsets:      true,
+		DeleteEmptydirData:    true,
+		Timeout:               5 * time.Minute,
+		EvictionTimeout:       1 * time.Minute,
```
ader1990: The eviction timeout should be in the order of magnitude of tens of minutes for dense storage related workloads, especially if Rook Ceph or Longhorn is running in the cluster. Even for external CSIs, it should be good to have a reasonable timeout.

test/e2e/node_scale_test.go
```diff
@@ -89,8 +89,8 @@ var _ = Describe("Workload cluster scaling", func() {
 					Namespace:                namespace.Name,
 					ClusterName:              clusterName,
 					KubernetesVersion:        e2eConfig.GetVariable(KubernetesVersion),
-					ControlPlaneMachineCount: ptr.To(int64(1)),
-					WorkerMachineCount:       ptr.To(int64(1)),
+					ControlPlaneMachineCount: ptr.To(int64(3)),
```
ader1990: To be reverted to their original values.

What is very interesting for me at first was the actual snap-k8s design for Control Planes - only 3 CPs participate to quorum, the rest are just back-up spares. Which probably derives from the raft specs? Or it was a snap-k8s decision?

pkg/cloudinit/scripts/install.sh
```diff
@@ -32,6 +32,14 @@ retry_snap_install() {
   return 1
 }
 
+# remove this
+  systemctl restart systemd-journald
```
ader1990, line 37: to be reverted.

PR #239, opened 2026-08-12
Upgrade capi version to 1.10.10

ader1990:
This PR does step by step upgrade for the capi version from 1.9.6 to the latest 1.10.x (1.10.10).

1.10.10 is the last capi release that was still backwards compatible.

This PR is part of the effort for: https://github.com/canonical/cluster-api-k8s/issues/208

There will be a subsequent PR to upgrade to the backwards incompatible version of capi >= 1.11.0

This PR addresses some python tox formatting rules that are now enabled by default, and a set of new properties needed for `framework.DumpAllResources(ctx, framework.DumpAllResourcesInput{		KubeConfigPath: input.ClusterProxy.GetKubeconfigPath(),		ClusterctlConfigPath: input.ClusterctlConfigPath`. Also, `e2eConfig.GetVariable` was changed to `e2eConfig.GetVariableOrEmpty`, as `e2eConfig.GetVariable` got removed.


ader1990:
PR is ready to be reviewed, all e2e tests pass. This PR does not have any business logic change, but all of the changes are needed for the upcoming PRs for the cluster api version upgrades.

claudiubelu:
LGTM. I'm not sure what the merge policy is here; if it's squash + merge, this can be ignored. But if it's rebase and merge, then IMO, a few things can be squashed:

- 6c35f2c0344e2c137b89162c7e0b4cb84bc38af8 and 38cd34d2e43c607523a0d25efa4ee48211453c28 - both related to upgrade to 1.9.7, one upgrades cluster-api, and the other cluster-api/test. The next commit already does this anyways (upgrade to 1.9.11)
- same with commits ed16f97f97ab25b21eb2f6ccd36127109356871a and 26c8f217310fbf704e4b9a61606b51a427eb7eed

Other than that:

- The changes in 25b0fd82f9fbbb2158c166780f1bb78b96ffcd04 (e2e: fix the new API breaking changes) are straight-forward, mostly using `GetVariableOrEmpty` instead of `GetVariable`.
- 50a32a0e3b00917f9321394670ebdd073a74d5d5: mostly updates the python typing style to a more modern version. One thing I disagree with, is the fact that BLE001 is ignored globally. That rule refers to general catchall `except Exception` blocks, which is not a good code pattern to use. If there are such cases, they should be handled properly, and if there really is one or a few cases in which this is fine, a `  # noqa: BLE001` can be put at the end of that line instead.
- 6f71db97db9e7c0f0b9f43315071e6aafac58c84: only adds `KubeConfigPath` and `ClusterctlConfigPath`. Other than that, only indentation noise.

ader1990:
Hello, would it be possible to merge this PR in preparation for the capi v1beta2 upgrade PRs?

cc @bschimke95 @HomayoonAlimohammadi @ktsakalozos 

Thanks.

estelacarmona:
Tagging also @deuberger so that he's aware

test/e2e/common.go
```diff
@@ -105,8 +108,8 @@ func dumpSpecResourcesAndCleanup(ctx context.Context, input cleanupInput) {
 	// that cluster variable is not set even if the cluster exists, so we are calling DeleteAllClustersAndWait
 	// instead of DeleteClusterAndWait
 	framework.DeleteAllClustersAndWait(ctx, framework.DeleteAllClustersAndWaitInput{
-		Client:    input.ClusterProxy.GetClient(),
-		Namespace: input.Namespace.Name,
+		ClusterProxy: input.ClusterProxy,
```
HomayoonAlimohammadi: `ClusterctlConfigPath` is missing here. In v1.10.10, `DeleteAllClustersAndWait` forwards it to `WaitForClusterDeleted`, which asserts `Expect(input.ClusterctlConfigPath).NotTo(BeEmpty())` — so this fails in `AfterEach` as soon as there is a cluster to delete.

CI does not catch it: `.github/workflows/e2e.yaml` runs with `SKIP_RESOURCE_CLEANUP=true`, which returns from `dumpSpecResourcesAndCleanup` at line 102 before reaching this call. Any local run with cleanup enabled breaks.

`input.ClusterctlConfigPath` is already on `cleanupInput`, so:

```go
framework.DeleteAllClustersAndWait(ctx, framework.DeleteAllClustersAndWaitInput{
    ClusterProxy:         input.ClusterProxy,
    ClusterctlConfigPath: input.ClusterctlConfigPath,
    Namespace:            input.Namespace.Name,
    ArtifactFolder:       input.ArtifactFolder,
}, input.IntervalsGetter(input.SpecName, "wait-delete-cluster")...)
```

`ArtifactFolder` is optional but enables the dump-on-deletion-timeout path.

test/e2e/cluster_upgrade.go
```diff
@@ -163,7 +164,7 @@ func ClusterUpgradeSpec(ctx context.Context, inputGetter func() ClusterUpgradeSp
 				Flavor:                   flavor,
 				Namespace:                namespace.Name,
 				ClusterName:              clusterName,
-				KubernetesVersion:        input.E2EConfig.GetVariable(KubernetesVersion),
+				KubernetesVersion:        input.E2EConfig.GetVariableOrEmpty(KubernetesVersion),
```
HomayoonAlimohammadi: `GetVariableOrEmpty` is the wrong replacement for the removed `GetVariable`. In `test@v1.10.10/framework/clusterctl/e2e_config.go`, `MustGetVariable` is a byte-for-byte rename of the old `GetVariable` (it keeps the `Expect(ok).To(BeTrue())`), whereas `GetVariableOrEmpty` is documented as "does not fail" and returns `""`.

So every converted call site loses its fail-fast check: a missing `KUBERNETES_VERSION` now renders a cluster template with an empty version and fails somewhere far downstream instead of at the config check. Upstream CAPI migrated its own call sites to `MustGetVariable`.

Applies to all 12 sites — here (167, 183, 191, 201), `e2e_suite_test.go` (245, 248), and `create_test.go`, `in_place_upgrade_test.go`, `node_scale_test.go`, `md_remediation_test.go`, `intermediate_ca_test.go`, `refresh_certs_test.go`, `orchestrated_*_in_place_upgrade_test.go`.

hack/release_promotion/tox.ini
```diff
@@ -41,7 +41,7 @@ deps =
 commands =
     isort {[vars]all_path}
     ruff format {[vars]all_path}
-    ruff check --fix {[vars]all_path}
+    ruff check --unsafe-fixes --ignore BLE001 --fix {[vars]all_path}
```
HomayoonAlimohammadi, line 44: Two concerns, both non-blocking:

- `--ignore BLE001` is added only to `format`, not to `lint` (line 60, `ruff check --diff`), so the two envs can now disagree. There is no `ruff.toml`/`pyproject.toml` anywhere in the repo, so ruff runs with its default select (`E4,E7,E9,F`) — `BLE` is not in that set, which makes this flag look like a no-op. A shared config file is the right place for an ignore.
- `--unsafe-fixes` on an auto-`--fix` command silently applies fixes ruff itself flags as semantics-changing.

hack/release_promotion/tests/test_promote.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/python
```
HomayoonAlimohammadi: `/usr/bin/python` does not exist on Ubuntu (only `python3`). `scripts/promote.py` uses `#!/usr/bin/env python3` — match that.

go.mod
```diff
@@ -7,41 +7,50 @@ require (
...
-	k8s.io/api v0.31.3
-	k8s.io/apimachinery v0.31.3
-	k8s.io/apiserver v0.31.3
-	k8s.io/client-go v0.31.3
+	k8s.io/api v0.32.3
+	k8s.io/apimachinery v0.32.3
+	k8s.io/apiserver v0.32.3
+	k8s.io/client-go v0.32.3
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/kubernetes v1.31.3
```
HomayoonAlimohammadi: Left at 1.31 while every other `k8s.io/*` moved to v0.32.3. It compiles, and the only usage is `podv1` in `pkg/ck8s/workload_cluster.go`, but the skew is worth closing — either bump to v1.32.3 or drop the dependency.

HomayoonAlimohammadi (commented):
> [!WARNING]
> **This review is AI generated.** It was produced by an AI assistant and may contain mistakes or miss context. Treat every finding as a suggestion to verify, not as ground truth.

Verified locally on `6f71db9`: `go build ./...`, `go vet ./...` and `go vet -tags e2e ./test/...` are clean, unit tests pass (envtest suites skipped — no local etcd binary), and the CRD regeneration is complete (no stale `Last time the condition transitioned` or `Map of string keys and values` text left in any manifest). All CI checks are green.

The dependency bump and the CRD regen look correct. Two findings worth addressing:

1. **`GetVariableOrEmpty` is the wrong replacement for the removed `GetVariable`** — `MustGetVariable` is the byte-for-byte rename that preserves the fail-fast assertion. As written, a missing variable silently becomes `""`. Affects all 12 converted call sites.
2. **`DeleteAllClustersAndWait` is missing `ClusterctlConfigPath`** in `test/e2e/common.go` — `WaitForClusterDeleted` asserts it is non-empty. CI passes only because e2e runs with `SKIP_RESOURCE_CLEANUP=true`, which returns before that call.

The rest (tox ruff flags, shebang, `k8s.io/kubernetes` version skew) is minor and non-blocking.

hack/release_promotion/tox.ini
ader1990: For some reason, BLE is now in the set. In any case, these errors just poped out without any actual reason, other than the newer tools to check are now more hardened. The Python code works correctly as is.

test/e2e/common.go
ader1990: updated.

go.mod
ader1990: updated.

HomayoonAlimohammadi (approved):

PR #240, opened 2026-08-12
Upgrade capi version to 1.13.4

ader1990:
This change focuses on using the newly backwards incompatible cluster api version 1.13.4. Since 1.11, capi has been deprecating the v1beta1 CRDs and controllers. The new v1betav2 capi CRD versions/libraries are used.

What works - E2E tests:

          - "Workload cluster creation"
          - "Workload cluster scaling"
          - "Certificate Refresh"
          - "KCP remediation"
          - "MachineDeployment remediation"
          - "Workload cluster upgrade"
          - "Orchestrated In place upgrades"

Pending work:
 * Discussion on wether to keep the betav1 conditions and statuses


ader1990:
Major logical changes:

  * Usage of capi apis v1beta2 using capi libraries v1.13.4
  * Added v1beta3 version for all ck8s* CRDs
  * Usage of the new v1beta3 version in the controllers
  * Updated v1beta2 version with the new imports
  * Added v1beta2 to v1beta3 conversions

Major implementation details:

  * CK8sControlPlane uses the new conditions from the v1beta2 capi contract
  * CK8sControlPlane has a new capi contract status property - required by the new capi contract

test/e2e/config/ck8s-incus.yaml
```diff
@@ -15,8 +15,8 @@ providers:
   - name: cluster-api
     type: CoreProvider
     versions:
-      - name: v1.9.6
-        value: https://github.com/kubernetes-sigs/cluster-api/releases/download/v1.9.6/core-components.yaml
+      - name: v1.13.4
+        value: https://github.com/ader1990/cluster-api/releases/download/v1.13.4/core-components.yaml
```
HomayoonAlimohammadi: Points CAPI core components at a personal fork. Must be `kubernetes-sigs/cluster-api` before merge.

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -175,16 +166,26 @@ func (r *CK8sConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	switch {
 	// Wait for the infrastructure to be ready.
-	case !cluster.Status.InfrastructureReady:
+	case !*cluster.Status.Initialization.InfrastructureProvisioned:
```
HomayoonAlimohammadi: Nil deref. `InfrastructureProvisioned` is `*bool` with `omitempty` and is nil until the infra provider sets it — this panics on a fresh cluster. Use `ptr.Deref(...)` as the rest of the PR does.

bootstrap/controllers/ck8sconfig_controller.go
```diff
@@ -193,7 +194,7 @@ func (r *CK8sConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 
 	// Note: can't use IsFalse here because we need to handle the absence of the condition as well as false.
-	if !conditions.IsTrue(cluster, clusterv1.ControlPlaneInitializedCondition) {
+	if !conditions.IsTrue(cluster, clusterv1.ClusterControlPlaneAvailableCondition) {
```
HomayoonAlimohammadi: Bootstrap needs the control plane *initialized* (API reachable so nodes can join), not *available*. Should be `ClusterControlPlaneInitializedCondition`.

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -74,28 +74,34 @@ type CK8sControlPlaneReconciler struct {
...
 	logger := r.Log.WithValues("namespace", req.Namespace, "ck8sControlPlane", req.Name)
 
+	logger.Info("CK8sControlPlaneReconciler reconcile request received")
+
 	// Fetch the CK8sControlPlane instance.
 	kcp := &controlplanev1.CK8sControlPlane{}
 	if err := r.Get(ctx, req.NamespacedName, kcp); err != nil {
 		if apierrors.IsNotFound(err) {
-			return reconcile.Result{}, nil
+			logger.Error(err, "Failed to retrieve CK8sControlPlane: Not Found")
```
HomayoonAlimohammadi, line 86: Returning the error on `NotFound` puts a deleted KCP into permanent exponential backoff with error logs. Should stay `return ctrl.Result{}, nil`.

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -122,52 +129,55 @@ func (r *CK8sControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		var connFailure *ck8s.RemoteClusterConnectionError
-		if errors.As(updateErr, &connFailure) {
-			logger.Info("Could not connect to workload cluster to fetch status", "updateErr", updateErr.Error())
-		} else {
-			logger.Error(updateErr, "Failed to update CK8sControlPlane Status")
-			err = kerrors.NewAggregate([]error{err, updateErr})
+		// Always attempt to Patch the CK8sControlPlane object and status after each reconciliation.
+		if patchErr := patchCK8sControlPlane(ctx, patchHelper, kcp); patchErr != nil {
+			logger.Error(patchErr, "Failed to patch CK8sControlPlane")
+			err = kerrors.NewAggregate([]error{err, patchErr})
```
HomayoonAlimohammadi: The defer aggregates into `err` (the local from `util.GetOwnerCluster` on L94), not the named return `reterr`. So `updateStatus` and `patchCK8sControlPlane` failures are logged and then silently dropped, and the `if err == nil` requeue guard on L165 always reads nil. Assign to `reterr`.

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -122,52 +129,55 @@ func (r *CK8sControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		if err == nil && !res.Requeue && res.RequeueAfter <= 0 && kcp.DeletionTimestamp.IsZero() {
+			logger.Info("Checking if to requeueing CK8sControlPlane for not ready status")
 
-	// TODO: remove this as soon as we have a proper remote cluster cache in place.
-	// Make KCP to requeue in case status is not ready, so we can check for node status without waiting for a full resync (by default 10 minutes).
-	// Only requeue if we are not going in exponential backoff due to error, or if we are not already re-queueing, or if the object has a deletion timestamp.
-	if err == nil && !res.Requeue && res.RequeueAfter <= 0 && kcp.DeletionTimestamp.IsZero() {
-		if !kcp.Status.Ready {
-			res = ctrl.Result{RequeueAfter: 20 * time.Second}
+			if !conditions.IsTrue(cluster, clusterv1.ClusterControlPlaneAvailableCondition) {
```
HomayoonAlimohammadi: Reads a mirror condition off the `cluster` object fetched at the top of Reconcile, so it lags the KCP's own state by a reconcile. Should read `AvailableCondition` off `kcp`.

controlplane/controllers/ck8scontrolplane_controller.go
```diff
@@ -104,7 +110,8 @@ func (r *CK8sControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	}
 
 	// Wait for the cluster infrastructure to be ready before creating machines
-	if !cluster.Status.InfrastructureReady {
+	if !conditions.IsTrue(cluster, clusterv1.InfrastructureReadyCondition) {
```
HomayoonAlimohammadi: `cluster.Status.InfrastructureReady` was a latch; the `InfrastructureReady` condition mirrors the infra provider and can flip back to False. The v1beta2 equivalent of the old semantics is `Status.Initialization.InfrastructureProvisioned`.

controlplane/controllers/machine_controller.go
```diff
@@ -69,7 +69,7 @@ func (r *MachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// if machine registered PreTerminate hook, wait for capi asks to resolve PreTerminateDeleteHook
 	if annotations.HasWithPrefix(clusterv1.PreTerminateDeleteHookAnnotationPrefix, m.Annotations) &&
 		m.Annotations[clusterv1.PreTerminateDeleteHookAnnotationPrefix] == ck8sHookName {
-		if !conditions.IsFalse(m, clusterv1.PreTerminateDeleteHookSucceededCondition) {
+		if conditions.Get(m, PreTerminateHookCleanupAnnotation) == nil {
```
HomayoonAlimohammadi: `PreTerminateHookCleanupAnnotation` is an annotation key (`pre-terminate.delete.hook.machine.cluster.x-k8s.io/ck8s`) being passed as a condition *type*. It will never match, so this always returns early and the ck8s cleanup never runs — machines stick in deletion. Was `!conditions.IsFalse(m, clusterv1.PreTerminateDeleteHookSucceededCondition)`.

pkg/ck8s/workload_cluster.go
```diff
@@ -603,25 +623,51 @@ func (w *Workload) UpdateAgentConditions(ctx context.Context, controlPlane *Cont
...
+					Reason:  clusterv1.MachineReadyReason,
+					Message: "",
+				})
+				conditions.Set(machine, metav1.Condition{
+					Type:    clusterv1.MachineUpToDateCondition,
+					Status:  metav1.ConditionTrue,
+					Reason:  clusterv1.MachineUpToDateReason,
+					Message: "",
+				})
 			}
```
HomayoonAlimohammadi, line 663: These three are CAPI-owned Machine conditions — `Ready`/`Available` are computed by the Machine controller and `UpToDate` by the owning controller. Setting them here means the provider and CAPI overwrite each other every reconcile, and `UpToDate` is asserted purely from node readiness with no reference to whether the machine spec matches the KCP spec. Combined with `scale.go` preflight now gating on exactly these, preflight is validating conditions this function fabricates. Keep `AgentHealthy` for preflight and let CAPI own these.

pkg/ck8s/workload_cluster.go
```diff
@@ -700,31 +750,56 @@ func aggregateFromMachinesToKCP(input aggregateFromMachinesToKCPInput) {
 		input.kcpErrors = append(input.kcpErrors, fmt.Sprintf("Following machines are reporting %s errors: %s", input.note, strings.Join(kcpMachinesWithErrors.List(), ", ")))
 	}
 	if len(input.kcpErrors) > 0 {
-		conditions.MarkFalse(input.controlPlane.KCP, input.condition, input.unhealthyReason, clusterv1.ConditionSeverityError, "%s", strings.Join(input.kcpErrors, "; "))
+		conditions.Set(input.controlPlane.KCP, metav1.Condition{
+			Type:    input.condition,
+			Status:  metav1.ConditionFalse,
+			Reason:  controlplanev1.GroupVersion.Version,
```
HomayoonAlimohammadi: `Reason` is set to the literal string `"v1beta3"` here and in the four branches below, leaving `unhealthyReason`/`unknownReason` as dead struct fields. Also in this function: the info branch now reports `ConditionTrue` where it previously reported False, and `kcpMachinesWithInfo` is never populated since the typed severity switch was replaced by a `strings.Contains(reason, "error"|"fail")` heuristic.

controlplane/controllers/scale.go
```diff
@@ -172,8 +173,11 @@ func (r *CK8sControlPlaneReconciler) preflightChecks(_ context.Context, controlP
 	}
 
 	// Check machine health conditions; if there are conditions with False or Unknown, then wait.
-	allMachineHealthConditions := []clusterv1.ConditionType{controlplanev1.MachineAgentHealthyCondition}
-
+	allMachineHealthConditions := []string{
+		clusterv1.MachineReadyCondition,
+		clusterv1.MachineAvailableCondition,
+		clusterv1.MachineUpToDateCondition,
+	}
```
HomayoonAlimohammadi, line 180: This drops `MachineAgentHealthyCondition`, so k8sd agent health no longer gates scale up/down. It replaces it with conditions that `UpdateAgentConditions` itself sets from node readiness, making the check circular.

controlplane/controllers/scale.go
```diff
@@ -356,15 +372,12 @@ func (r *CK8sControlPlaneReconciler) generateMachine(ctx context.Context, kcp *c
...
 			InfrastructureRef: *infraRef,
 			Bootstrap: clusterv1.Bootstrap{
-				ConfigRef: bootstrapRef,
+				ConfigRef: *bootstrapRef,
 			},
-			FailureDomain:           failureDomain,
-			NodeDrainTimeout:        kcp.Spec.MachineTemplate.NodeDrainTimeout,
-			NodeVolumeDetachTimeout: kcp.Spec.MachineTemplate.NodeVolumeDetachTimeout,
-			NodeDeletionTimeout:     kcp.Spec.MachineTemplate.NodeDeletionTimeout,
+			FailureDomain: *failureDomain,
```
HomayoonAlimohammadi, line 380: `NodeDrainTimeout`, `NodeVolumeDetachTimeout` and `NodeDeletionTimeout` were dropped here, but `CK8sControlPlaneMachineTemplate` still declares all three and the e2e templates still set them — they are now silently ignored. In v1beta2 they moved to `MachineSpec.Deletion.NodeDrainTimeoutSeconds` etc. and need a Duration→seconds mapping.

controlplane/controllers/remediation.go
```diff
@@ -136,29 +137,29 @@ func (r *CK8sControlPlaneReconciler) reconcileUnhealthyMachines(ctx context.Cont
...
 		// Executes checks that apply only if the control plane is already initialized; in this case KCP can
 		// remediate only if it can safely assume that the operation preserves the operation state of the
 		// existing cluster (or at least it doesn't make it worse).
 
 		// The cluster MUST have more than one replica, because this is the smallest cluster size that allows any etcd failure tolerance.
 		if controlPlane.Machines.Len() <= 1 {
 			log.Info("A control plane machine needs remediation, but the number of current replicas is less or equal to 1. Skipping remediation", "Replicas", controlPlane.Machines.Len())
-			conditions.MarkFalse(machineToBeRemediated, clusterv1.MachineOwnerRemediatedCondition, clusterv1.WaitingForRemediationReason, clusterv1.ConditionSeverityWarning, "KCP can't remediate if current replicas are less or equal to 1")
+			v1beta1conditions.MarkFalse(machineToBeRemediated, clusterv1.MachineOwnerRemediatedCondition, clusterv1.RemediationInProgressV1Beta1Reason, clusterv1.ConditionSeverityWarning, "KCP can't remediate if current replicas are less or equal to 1")
```
HomayoonAlimohammadi, line 148: Reason is inverted: this is a *blocked* path but reports `RemediationInProgress`. Same on L155, L162, L288, L297 — should be `WaitingForRemediationV1Beta1Reason`. Conversely L186 (delete failed) now reports `WaitingForRemediation` and should be `RemediationFailedV1Beta1Reason`. Looks like a bad bulk replace.

controlplane/controllers/remediation.go
```diff
@@ -61,8 +62,8 @@ func (r *CK8sControlPlaneReconciler) reconcileUnhealthyMachines(ctx context.Cont
 
 			conditions.Delete(m, clusterv1.MachineOwnerRemediatedCondition)
 
-			if err := patchHelper.Patch(ctx, m, patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{
-				clusterv1.MachineOwnerRemediatedCondition,
+			if err := patchHelper.Patch(ctx, m, patch.WithOwnedConditions{Conditions: []string{
+				string(clusterv1.MachineOwnerRemediatedCondition),
 			}}); err != nil {
 				errList = append(errList, fmt.Errorf("failed to patch machine %s: %w", m.Name, err))
```
HomayoonAlimohammadi, line 68: Mixed condition sets: this deletes the v1beta2 `MachineOwnerRemediated`, but everything below writes the deprecated v1beta1 one via `v1beta1conditions.MarkFalse` while declaring ownership through `WithOwnedConditions` (v1beta2, not `WithOwnedV1Beta1Conditions`). CAPI 1.13's MHC flow reads the v1beta2 condition, so it never sees the provider's remediation state. Pick one set.

metadata.yaml
```diff
@@ -18,4 +18,7 @@ releaseSeries:
     contract: v1beta1
   - major: 0
     minor: 6
-    contract: v1beta1
+    contract: v1beta2
+  - major: 0
+    minor: 7
+    contract: v1beta3
```
HomayoonAlimohammadi, line 24: `contract` is the CAPI contract version, not the provider's API version — there is no `v1beta3` CAPI contract. Should be `v1beta2`, matching `controlplane/metadata.yaml`.

.golangci.yml
```diff
@@ -135,6 +135,22 @@ linters:
...
+        path: controlplane
+        text: SA1019
+      - linters:
+          - gosec
+        path: bootstrap
+        text: G117
+      - linters:
+          - gosec
+        path: pkg
+        text: G101
```
HomayoonAlimohammadi, line 153: Blanket `SA1019` across all of `bootstrap/` and `controlplane/` hides exactly the deprecation warnings this migration exists to surface. `G101` across all of `pkg/` also replaces a single-line `//nolint:gosec` that was removed from `workload_cluster.go`. Prefer targeted nolints.

controlplane/api/v1beta2/conversion.go
```diff
@@ -16,8 +16,101 @@ limitations under the License.
...
+	if len(src.Status.Conditions) > 0 {
+		if err := roundTripConvert(src.Status.Conditions, &dst.Status.Conditions); err != nil {
+			return fmt.Errorf("failed to convert conditions from v1beta3: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func roundTripConvert(src any, dst any) error {
```
HomayoonAlimohammadi, line 110: JSON round-trip conversion is lossy by construction and silent — any field whose shape changed between v1beta2 and v1beta3 is dropped with no error. CAPI's convention is conversion-gen plus a restore annotation for spoke-only fields. Also note v1beta1 conditions allow an empty `Reason`, but `metav1.Condition.Reason` is `MinLength=1` with a pattern, so converting an existing stored object with an empty reason produces a body the apiserver rejects. Worth a round-trip fuzz test before this ships.

HomayoonAlimohammadi (changes requested):
> **Note:** This review was generated by AI (Claude). Findings were verified against the CAPI v1.13.4 source, but please sanity-check before acting on them.

Reviewed the full diff against `main`. It builds and unit tests pass. The bulk of the v1beta1→v1beta2 port and the new v1beta3 provider API look correct, but several find-and-replace conversions changed behavior in ways that appear unintended, and the e2e config points at a personal fork.

**Blockers**

- `test/e2e/config/ck8s-incus.yaml` pulls CAPI core components from `github.com/ader1990/cluster-api` instead of `kubernetes-sigs`.
- Nil deref in `ck8sconfig_controller.go` on `*cluster.Status.Initialization.InfrastructureProvisioned`.
- `CK8sControlPlaneReconciler.Reconcile` aggregates deferred errors into a local `err` rather than the named return `reterr`, so status/patch failures are silently swallowed.
- `NotFound` on the KCP now returns an error, putting deleted objects into permanent backoff.
- The PreTerminate hook check passes an annotation key as a condition type, so the ck8s cleanup never runs.
- `UpdateAgentConditions` writes CAPI-owned Machine conditions (`Ready`/`Available`/`UpToDate`), and preflight then gates on those same fabricated conditions — the `AgentHealthy` gate is gone.
- Remediation writes deprecated v1beta1 conditions while declaring v1beta2 ownership; CAPI's MHC flow won't see them. The reason constants are also inverted between the "waiting" and "failed" paths.
- `MachineTemplate` drain/detach/deletion timeouts are no longer propagated to the Machine, though the API and e2e templates still set them.
- `metadata.yaml` declares `contract: v1beta3`; there is no such CAPI contract.

**Worth addressing**

- `ck8sconfig_controller.go` gates worker joins on `ControlPlaneAvailable` rather than `ControlPlaneInitialized`.
- `cluster.Status.InfrastructureReady` (a latch) was replaced with a mirrored condition that can flip back to False.
- `aggregateFromMachinesToKCP` sets every condition `Reason` to the literal `"v1beta3"`, inverts the info branch to True, and replaces typed severity with a `strings.Contains` heuristic on the reason.
- `conditions.SetAggregate`/`SetSummary` were dropped as "not available" — the v1beta2 equivalents (`NewAggregateCondition`, `SetSummaryCondition`) do exist, and `MachinesReadyCondition` is now never set despite still being listed in `WithOwnedConditions`.
- Several condition messages dropped `err.Error()`, losing the cause.
- Blanket `SA1019`/`G101` lint suppressions across whole trees.
- `go.mod` downgrades `k8s.io/kubernetes` v1.32.3 → v1.31.3 while every other k8s lib moves to v0.35.4, and pins `go 1.26.7` (patch-level) as a hard floor.
- The JSON round-trip conversion webhooks are silently lossy; a round-trip fuzz test would be worthwhile.

**Minor**: `getGetFailureDomainIDs` typo; duplicate `AgentHealthy` constants in the same package; `NextFailureDomainForScaleUp` returns an `error` it never populates and both callers discard it; `pkgerrors` introduced for one wrap in a `fmt.Errorf` file; unused `ctx` params on `setReplicas`/`setInitializedCondition`; `kcp.Status.Replicas` assigned twice; `NewMachine` is dead code that now dereferences `*failureDomain` unguarded; leftover `metadata.labels.key: value` in the e2e template; the Cluster watch predicate was dropped rather than ported to `ClusterPausedTransitionsOrInfrastructureProvisioned`.

The "keep v1beta1 conditions and statuses" question listed as pending in the description probably wants settling first, since the remediation path currently straddles both.

test/e2e/config/ck8s-incus.yaml
ader1990: Will be updated. This is a dev fork I have used to add more logs in critical capi places.

bootstrap/controllers/ck8sconfig_controller.go
ader1990: Actually, the ControlPlaneInitializedCondition is a lot less strong - the ControlPlaneInitializedCondition is a condition that does not make any assumption that the cluster is actually available. And  the ck8sconfig controller needs to access the k8sd api node directly which means that the cluster needs in itself to be available.

I will change it as per the review request back to ClusterControlPlaneInitializedCondition, but wanted to leave the above not as future reference, in case we need a stronger condition.

controlplane/controllers/ck8scontrolplane_controller.go
ader1990: This is a leftover from the initial iteration that looked for the controlplane in the default namespace. Will be updated.

controlplane/controllers/machine_controller.go
ader1990: Hmm, the cleanup always works. this means that the whole logic is either redundant or wrong.
Changed to the suggested version, but I will investigate more on this.

metadata.yaml
ader1990: interesting, the 0.6 version was using v1beta2, but it actually used v1beta1 for some reason.

controlplane/controllers/remediation.go
ader1990: The remediation e2e test case works though. Will remove the v1beta1 conditions.

metadata.yaml
ader1990: if I change it -> clusterctl bootstrap fails with `msg: "failed to find releases tagged with a valid semantic version number"`.

ader1990: reverted the change back to the original.

controlplane/controllers/machine_controller.go
```diff
@@ -49,60 +60,54 @@ func (r *MachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manag
...
-		// have transferred their leadership role (if they were the leader).
-		//
-		// In the case of Canonical Kubernetes, the node removal happens by executing the k8sd RemoveNode endpoint, which takes care of
-		// removing the node from the datastore quorum as well. Therefore, we should not need to do any more actions in this case. It should
-		// suffice to simply delete the annotation.
-		//
-		// Note that this currently makes the annotation a no-op in the code here. However, we still keep the logic in the code is case it
-		// is needed in the future.
-
+		logger.Info("removing the annotation PreTerminateDeleteHookAnnotationPrefix")
```
ader1990, line 94: here, before removing the annotation, we need to make sure that the node is cleaned up and the underlying machine can be wiped.

first, we need to call https://github.com/canonical/k8s-snap-api/blob/c23486a68644d6a2a381f9c731d26e1b2c5bf513/api/rpc_node_status.go#L11 - get node status - to make sure that the node is actually gone.

if the node has status, we need to issue a removeNode Request and trigger a new reconcile after a few seconds.

if the node cannot be found - aka status check fails, we can safely remove the annotation.

Note that reaching this state, the capi controller already cleaned up the node in terms of workloads and volumes: the node was already drained, volumes were detached.

controlplane/controllers/scale.go
ader1990: Readded these configurations.

ader1990:
**Blockers**

- DONE. `test/e2e/config/ck8s-incus.yaml` pulls CAPI core components from `github.com/ader1990/cluster-api` instead of `kubernetes-sigs`.
- DONE. Nil deref in `ck8sconfig_controller.go` on `*cluster.Status.Initialization.InfrastructureProvisioned`.
- DONE. `CK8sControlPlaneReconciler.Reconcile` aggregates deferred errors into a local `err` rather than the named return `reterr`, so status/patch failures are silently swallowed.
- DONE. `NotFound` on the KCP now returns an error, putting deleted objects into permanent backoff.
- DONE. The PreTerminate hook check passes an annotation key as a condition type, so the ck8s cleanup never runs.
- TBD. `UpdateAgentConditions` writes CAPI-owned Machine conditions (`Ready`/`Available`/`UpToDate`), and preflight then gates on those same fabricated conditions — the `AgentHealthy` gate is gone.
- TBD. Remediation writes deprecated v1beta1 conditions while declaring v1beta2 ownership; CAPI's MHC flow won't see them. The reason constants are also inverted between the "waiting" and "failed" paths.
- DONE. `MachineTemplate` drain/detach/deletion timeouts are no longer propagated to the Machine, though the API and e2e templates still set them.
- NOT APPLIABLE. `metadata.yaml` declares `contract: v1beta3`; there is no such CAPI contract.

**Worth addressing**

- DONE. `ck8sconfig_controller.go` gates worker joins on `ControlPlaneAvailable` rather than `ControlPlaneInitialized`.
- DONE. `cluster.Status.InfrastructureReady` (a latch) was replaced with a mirrored condition that can flip back to False.
- DONE. `aggregateFromMachinesToKCP` sets every condition `Reason` to the literal `"v1beta3"`, inverts the info branch to True, and replaces typed severity with a `strings.Contains` heuristic on the reason.
- TBD. `conditions.SetAggregate`/`SetSummary` were dropped as "not available" — the v1beta2 equivalents (`NewAggregateCondition`, `SetSummaryCondition`) do exist, and `MachinesReadyCondition` is now never set despite still being listed in `WithOwnedConditions`.
- To be investigated. Several condition messages dropped `err.Error()`, losing the cause.
- TBD. Blanket `SA1019`/`G101` lint suppressions across whole trees.
- TBD. `go.mod` downgrades `k8s.io/kubernetes` v1.32.3 → v1.31.3 while every other k8s lib moves to v0.35.4, and pins `go 1.26.7` (patch-level) as a hard floor.
- TBD. The JSON round-trip conversion webhooks are silently lossy; a round-trip fuzz test would be worthwhile.

**Minor**: `getGetFailureDomainIDs` typo; duplicate `AgentHealthy` constants in the same package; `NextFailureDomainForScaleUp` returns an `error` it never populates and both callers discard it; `pkgerrors` introduced for one wrap in a `fmt.Errorf` file; unused `ctx` params on `setReplicas`/`setInitializedCondition`; `kcp.Status.Replicas` assigned twice; `NewMachine` is dead code that now dereferences `*failureDomain` unguarded; leftover `metadata.labels.key: value` in the e2e template; the Cluster watch predicate was dropped rather than ported to `ClusterPausedTransitionsOrInfrastructureProvisioned`.

It seems that the v1beta1 conditions are here to stay, as long as they are used also in the capi upstream -> The "keep v1beta1 conditions and statuses" question listed as pending in the description probably wants settling first, since the remediation path currently straddles both.

